PR #11, opened 2022-08-12
build: Change dockerfile to use intels fortran compiler

mariuswinger:
Changed Dockerfile and and compile.sh to use Intel's Fortran compiler through oneAPI. Building the docker image takes a bit of time. Everything is very slow on M1 Mac when using the platform flag in docker-compose. This is expected, but testing on other platforms should be done to see if this is a local problem or connected to the use of Fortran compiler. 

Note that the tests will not pass as long as dat/Asg_20VA001.dat and info/CompEXP.inf (and possibly the other files inside info) is present inside the api folder. The files can be found in the gpa-libhg repo.
I am not yet sure whether or not these files are used for testing or by the generated library as well. Though I suspect that the Fortran functions calls on these files => They are necessary to run calculations. At this point in time both the dat and info folder is added to the gitignore, so we should consider adding them back in or possibly adding them to /usr/local/lib/, but I suspect the Fortran code would not find them there.


soofstad:
Tested on x86 linux. Build time is approx 2min, and tests take about 5 seconds to run. I think this is expected and acceptable.

api/Dockerfile
```diff
@@ -3,10 +3,24 @@ ARG LIBHG_REPO_DEPLOY_KEY
...
 
-# Update and install Fortran compiler and f2py(Fortan to python, included in numpy)
+# Update and install f2py(Fortan to python, included in numpy)
 RUN apt-get update && \
-    apt-get install gfortran git -y && \
-    pip install numpy=="1.23.0"
+    DEBIAN_FRONTEND=noninteractive apt-get install -y --no-install-recommends \
+    curl ca-certificates gnupg2 git openssh-client && \
+    pip install numpy=="1.23.0" && \
+    rm -rf /var/lib/apt/lists/*
```
soofstad, line 11:
```suggestion
```
The directory is already empty

api/Dockerfile
```diff
@@ -3,10 +3,24 @@ ARG LIBHG_REPO_DEPLOY_KEY
 ENV LANG=C.UTF-8 LC_ALL=C.UTF-8
 WORKDIR /build
 
-# Update and install Fortran compiler and f2py(Fortan to python, included in numpy)
+# Update and install f2py(Fortan to python, included in numpy)
```
soofstad, line 6:
```suggestion
# Installing build dependencies. Including f2py(Fortan to python, included in numpy)
```

api/Dockerfile
```diff
@@ -3,10 +3,24 @@ ARG LIBHG_REPO_DEPLOY_KEY
...
 RUN apt-get update && \
-    apt-get install gfortran git -y && \
-    pip install numpy=="1.23.0"
+    DEBIAN_FRONTEND=noninteractive apt-get install -y --no-install-recommends \
```
soofstad, line 8:
```suggestion
    apt-get install -y --no-install-recommends \
```
These does not seem needed

api/Dockerfile
```diff
@@ -3,10 +3,24 @@ ARG LIBHG_REPO_DEPLOY_KEY
...
+    curl ca-certificates gnupg2 git openssh-client && \
+    pip install numpy=="1.23.0" && \
+    rm -rf /var/lib/apt/lists/*
+
+# Install Intel Fortran compiler
+RUN curl -fsSL https://apt.repos.intel.com/intel-gpg-keys/GPG-PUB-KEY-INTEL-SW-PRODUCTS-2023.PUB | gpg --dearmor > /usr/share/keyrings/oneAPI-keyring.gpg
+RUN echo "deb [signed-by=/usr/share/keyrings/oneAPI-keyring.gpg] https://apt.repos.intel.com/oneapi all main " > /etc/apt/sources.list.d/oneAPI.list
+RUN apt-get update && \
+    DEBIAN_FRONTEND=noninteractive apt-get install -y --no-install-recommends \
```
soofstad, line 17:
```suggestion
    apt-get install -y --no-install-recommends \
```

api/Dockerfile
```diff
@@ -3,10 +3,24 @@ ARG LIBHG_REPO_DEPLOY_KEY
...
+    build-essential openssh-client intel-oneapi-compiler-fortran && \
+    rm -rf /var/lib/apt/lists/*
+
+# Set environment variables
+ENV PATH="/opt/intel/oneapi/compiler/2022.1.0/linux/bin/intel64:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin" \
```
soofstad, line 22:
```suggestion
ENV PATH="/opt/intel/oneapi/compiler/2022.1.0/linux/bin/intel64:$PATH" \
```
Conventional way of appending directories to PATH

api/Dockerfile
```diff
@@ -3,10 +3,24 @@ ARG LIBHG_REPO_DEPLOY_KEY
...
+    LD_LIBRARY_PATH="/opt/intel/oneapi/compiler/2022.1.0/linux/lib:/opt/intel/oneapi/compiler/2022.1.0/linux/lib/x64:/opt/intel/oneapi/compiler/2022.1.0/linux/compiler/lib/intel64_lin"
 
 # Clone libhg fortran repo (cfortan compatible branch)
```
soofstad, line 25:
```suggestion
# Clone libhg fortran repo
```

soofstad: Also, no need for line 30 anymore

api/compile.sh
```diff
@@ -1,23 +1,23 @@
 #!/bin/bash
-set -eux
```
soofstad, line 2: This is always good to have.
For backgound, see; http://redsymbol.net/articles/unofficial-bash-strict-mode/

soofstad (changes requested):

soofstad (approved):
:ok_hand:

PR #13, opened 2022-08-17
feat: add api endpoints and tests

### LINKED ISSUE #3: API first: Try to define sensible endpoints

mariuswinger:
Add use cases and feature for Chemeq and Multiflash with tests. Also added changed classes to pydantic models, added validators and expanded descriptions on these classes.

Regarding test_chemeq and the last test (testing the computations) it sometimes outputs a single mole entry that is much greater than the others => test fail. By running the tests again (change nothing) it passes ... or fails. There is no logical explanation. This is a weird bug (This only seems to happen on my Intel chipped mac).

Tested on both M1 Mac and Intel Mac.

closes #3


api/src/entities/Chemeq.py
```diff
@@ -1,29 +1,88 @@
...
+        for row in v:
+            if not len(row) == number_of_columns:
+                raise ValueError(f"Stoichiometric matrix must have {number_of_columns} columns")
+            row_count += 1
+        if not row_count == number_of_rows:
+            raise ValueError(f"Stoichiometric matrix must have {number_of_rows} rows")
+        return v
 
-    def compute(self) -> Tuple[np.ndarray, np.ndarray, np.ndarray, np.ndarray]:
+    def compute(self) -> Tuple[npt.NDArray[bytes], npt.NDArray[float], npt.NDArray[float], npt.NDArray[float]]:
```
soofstad, line 85: Instead of using a basic tuple here, we should consider NamedTuple.
Usually much more pleasant to work with.
https://docs.python.org/3/library/collections.html#collections.namedtuple

api/src/entities/Chemeq.py
```diff
@@ -1,29 +1,88 @@
...
-class Chemeq:
-    number_of_components: int
-    component_ids: List[int]
-    number_of_reactions: int
-    feed_composition: List[float]
-    formula_matrix: List[List[int]]
-    stoichiometric_matrix: List[List[float]]
+class Chemeq(BaseModel):
+    """
+    Pydantic class which contains requests for computing simultaneous chemical and phase equilibria
```
soofstad, line 10:
```suggestion
    Model for the  simultaneous chemical and phase equilibria calculation inputs
```
Since this text is used for the API documentation. API consumers does not need to know that it is a pydantic class. But only what the generated schema is describing

api/src/entities/Multiflash.py
```diff
@@ -1,28 +1,58 @@
...
-@dataclass(frozen=True)
-class Multiflash:
-    number_of_components: int
-    component_ids: List[int]
-    temperature: float
-    pressure: float
-    composition: List[float]
+class Multiflash(BaseModel):
+    """
+    Pydantic class containing the request for computing multiphase flash.
```
soofstad, line 10:
```suggestion
    Model for the computing multiphase flash calculation input
```

api/src/entities/Multiflash.py
```diff
@@ -1,28 +1,58 @@
...
-        class_fields = {f.name for f in fields(cls)}
-        return Multiflash(**{k: v for k, v in dict_.items() if k in class_fields})
+    @validator("composition")
+    def validate_composition(cls, v, values):
+        if len(v) != values["number_of_components"]:
+            raise ValueError("Length of composition must match number_of_components.")
+        return v
 
-    def compute(self) -> Tuple[np.ndarray, np.ndarray, np.ndarray]:
+    def compute(self) -> Tuple[npt.NDArray[bytes], npt.NDArray[float], npt.NDArray[float]]:
```
soofstad, line 55: Same with namedTuple

soofstad (changes requested):

api/src/entities/Chemeq.py
mariuswinger: Nice! Makes the code a lot cleaner.

api/src/entities/Chemeq.py
mariuswinger: Yeah, good suggestion.

soofstad (approved):

PR #18, opened 2022-08-29
ci: add github actions CI

mariuswinger:


.github/workflows/on-tag.yaml
```diff
@@ -0,0 +1,46 @@
...
+        run: |
+          TAG_REF=$(echo $GITHUB_REF | cut -d / -f 3)
+          echo "::set-output name=tag-ref::TAG_REF"
+
+  publish-tag-ref:
+    needs: [set-tag-ref, tests]
+    uses: ./.github/workflows/publish.yaml
+    with:
+      image-tag: ${{ needs.set-tag-ref.outputs.tag-ref }}
+      oauth-redirect-url: REPLACE_ME
```
soofstad, line 38: IMO, there is no need to publish git-ref images for this project

.github/workflows/tests.yaml
```diff
@@ -0,0 +1,92 @@
+name: "Test"
+on:
+  # Workflow dispatch is used for manual triggers
+  workflow_dispatch:
+  # Workflow call is used for called from another workflow
+  workflow_call:
+    secrets:
+      CR_SECRET:
+        description: 'Secret to authenticate if using an other container registry than Github'
+        required: false
```
soofstad, line 10: Not needed here
```suggestion
```

soofstad (approved):

PR #21, opened 2022-08-31
feat: calculate fluid

### LINKED ISSUE #20: Web: Calculate component

olavis:
Reusable card component + input form for fluid calculations
Closes #20


web/src/components/common/Card.tsx
```diff
@@ -0,0 +1,33 @@
+import { Card as CardWrapper, Typography } from '@equinor/eds-core-react'
+import styled from 'styled-components'
+
+const StyledCard = styled(CardWrapper)`
+  min-width: 250px;
+  background-color: #f7f7f7;
+
+  .Typography__StyledTypography-sc-179guof-0 {
+    color: #007079;
+  }
```
soofstad, line 10: This is likely to break. Rather use styled components

web/src/components/feature/CalculateFluid.tsx
```diff
@@ -0,0 +1,64 @@
...
+const FluidPackage = styled.div`
+  display: flex;
+  gap: 16px;
+
+  .Button__ButtonBase-sc-1hs0myn-1 {
+    align-self: flex-end;
+  }
+
+  .Autocomplete__Container-sc-yvif0e-0 {
+    flex-grow: 1;
```
soofstad, line 25: Same

web/src/components/common/Card.tsx
```diff
@@ -0,0 +1,33 @@
+import { Card as CardWrapper, Typography } from '@equinor/eds-core-react'
+import styled from 'styled-components'
+
+const StyledCard = styled(CardWrapper)`
+  min-width: 250px;
+  background-color: #f7f7f7;
+
+  .Typography__StyledTypography-sc-179guof-0 {
+    color: #007079;
```
soofstad, line 9: Import colors from eds-tokens

soofstad (changes requested):

web/src/components/common/Card.tsx
olavis: Disagree. So far, we have no indications that this is likely to break. Nesting rules is even suggested by styled components to reduce the need for classes:

> nesting is a feature intentionally ported from Sass. Used sparingly it's a great way to lighten your code by reducing the need to create explicit classes for every element.
[https://styled-components.com/docs/faqs#can-i-nest-rules](url)

web/src/components/common/Card.tsx
olavis: #22

web/src/components/common/Card.tsx
soofstad: Nesting rules are no problem. But refering to css classes by styled-components could be, see; https://stackoverflow.com/questions/57993773/is-it-safe-to-target-styled-components-class-names

So, the className(hash) is calculated based on props and styles. If(when) EDS updates this, our web app will look weird. Not a big deal, there could be other issues with updating packages as well, but a good enough reason that this pattern should be avoided if possible.

soofstad (approved):

web/src/pages/Main.tsx
```diff
@@ -1,9 +1,12 @@
...
+import { Header } from '../components/common/Header'
 import { useContext } from 'react'
 import { AuthContext } from 'react-oauth2-code-pkce'
 
-export const MainPage = (): JSX.Element => {
-  const { token } = useContext(AuthContext)
+export const MainPage = (): JSX.Element | null => {
+  const { token, loginInProgress } = useContext(AuthContext)
+
+  if (loginInProgress) return null
```
olavis, line 8: Use conditional rendering instead of returning null component

olavis: @soofstad I reset your commit - it wasn't related to this issue (#20), so if you open a separate PR to review it, we won't have to await merging this one.

PR #26, opened 2022-09-01
feat: add components api endpoint

### LINKED ISSUE #17: API Endpoint: dictionary/hash map for component and component ids

mariuswinger:

This pull request is needed because components ID's need to accessible both for web and api.

Added a component_id dictionary to the common file in api/src. Created an endpoint which returns this.

closes #17


api/src/features/components/component_feature.py
```diff
@@ -0,0 +1,12 @@
+from typing import Dict
+
+from fastapi import APIRouter
+
+from common.components import COMPONENT_IDS
+
+router = APIRouter(tags=["component"])
+
+
```
soofstad, line 9: Not sure how this could fail, but should use the @create_response decorator anyway...

soofstad (approved):

api/src/features/components/component_feature.py
mariuswinger: Oops. WIll add.

soofstad (approved):

PR #28, opened 2022-09-07
feat: multiflash returns dictionary with component ids

### LINKED ISSUE #27: Result from multiflash should be a dict with [component_id]: value
Use a dict for result to correlate component with moles, instead of implicit index in array.

mariuswinger:

Multiflash returns a matrix of moles where the rows are in the order of the component ids (i.e. they are loosely coupled). A dictionary makes this coupling more apparent.

PR changes what multiflash returns (MultiflashResponse).

closes #27


soofstad:
This looks good. But could we give the input the same treatment? That is having components and their distribution coming in as a dict.
Would make a lot of this code simpler, and could be added in this PR.

olavis (approved):
Good 👍

soofstad (dismissed):
nice!

soofstad (approved):

PR #38, opened 2022-09-20
tests: add more tests

### LINKED ISSUE #34: Tests: Add new test data

mariuswinger:

More input/output has been supplied from Athens. Tests are updated with these.

Changes are made to Multiflash, MultiflashResponse and MultiflashResults. NamedTuples are introduced to handle the data in the back-end. The response is now a nested dictionary with the necessary data to display.

Closes #34


api/src/tests/conftest.py
```diff
@@ -24,32 +27,50 @@ def test_app():
...
+        elif config.getoption("--unit"):
+            if "unit" not in str(item.path):
+                item.add_marker(pytest.mark.skip(reason=f"--unit option is passed: {item.name} is not a unit test"))
 
 
 def read_json(filepath: str):
+    """Returns the contents of a json file at the filepath given"""
     with open(filepath, "r") as file:
         data = file.read()
     return json.loads(data)
```
soofstad, line 59: Ref talk about adding test data as pythond dict variables. No loading from file

api/src/tests/conftest.py
```diff
@@ -24,32 +27,50 @@ def test_app():
...
 
 
 @pytest.fixture(scope="module")
-def get_multiflash_test_data():
+def get_multiflash_test_data(request) -> (Dict, MultiflashResult):
```
soofstad, line 63: And this could be renamed to "convert_response_data_to_classes", or something....

soofstad (changes requested):

api/src/tests/conftest.py
mariuswinger: Went for something along those lines. The only "iffy" part is that the data from the output which is converted is not strictly the same as the response data (the response data does not contain mass concentrations). Not too sure if this is a minuscule non-issue or if the function should be renamed.

soofstad: Right. IMO the test data should not require any conversion. It should be the exact API response.
If we need additional test data for testing the library output, that should be a separate test with separate data.
Any good reasons why this is not the case?

mariuswinger: Reason (not exactly good): trying to be elegant.

soofstad (approved):
Få gjerne Olava til å se over også

api/src/common/utils/tuples.py
```diff
@@ -0,0 +1,54 @@
+from typing import Dict, NamedTuple
+
+import numpy as np
+import numpy.typing as npt
+
+from common.utils.enums import PhaseLabels
+
+
+class PhaseValues(NamedTuple):
+    """NamedTuple containing the results from Multiflash computation"""
```
olavis, line 10: Duplicated description

olavis (dismissed):

soofstad (dismissed):

soofstad (approved):

PR #39, opened 2022-09-23
Feat/mole result table

### LINKED ISSUE #23: Implement results table

soofstad:

Users should be able to see the mole results from the calculations

- Update generated TS api
- Fix a bug where local environment used wrong OAuth2 redirect URL
- Added the "MoleTable"
- Added tests for the "MoleTable"
- Enable yarn tests in the CI
- Removed App.test.tsx as a lot of mocking is required to test the AuthProvider

Screenshot_20220923_131142

closes #23


.pre-commit-config.yaml
```diff
@@ -60,6 +60,7 @@ repos:
       - id: isort
         name: isort (python)
         files: ^api/.*\.py$
+        args: ["--diff"]
```
mariuswinger, line 63: why?

api/src/entities/Multiflash.py
```diff
@@ -1,10 +1,11 @@
 from typing import Dict, NamedTuple
 
-import libhg
 import numpy as np
 import numpy.typing as npt
 from pydantic import BaseModel, Field, validator
 
+import libhg
```
mariuswinger, line 7: again, why?

web/src/api/MercuryAPI.tsx
```diff
@@ -1,9 +1,10 @@
-import { Configuration, TodosApi } from './generated'
+import { ComponentApi, Configuration } from './generated'
 
-export class MercuryAPI extends TodosApi {
+// TODO: Combine all generated api's into MercuryAPI
+export class MercuryAPI extends ComponentApi {
   constructor(token: string) {
     const configuration = new Configuration({
-      basePath: 'http://localhost',
```
mariuswinger, line 6: Not sure I understand what this does and why it is/is not needed?

soofstad: its used as a prefix for every api request.
If it's set to localhost it will be that in radix as well. Which wont work.

if its an empty string, it will use the "where i'm being served from", which will be correct

web/src/components/feature/MoleTable.test.tsx
```diff
@@ -0,0 +1,73 @@
+import { render, screen } from '@testing-library/react'
+import { MoleTable } from './MoleTable'
+import '@testing-library/jest-dom/extend-expect'
+
+const testComponents = {
+  '1': { chemical_formula: 'CO2', alt_name: 'Carbondioxide' },
```
mariuswinger, line 6: Should these be camelCase?

.pre-commit-config.yaml
soofstad: should not be included. Had some issues with isort in CI. Fixed now.

api/src/entities/Multiflash.py
soofstad: Issues with isort. Looks to be fixed now.

web/src/components/feature/MoleTable.test.tsx
soofstad: Yes, but that is what they look like at the moment coming from the API.

olavis:
> ## What does this pull request change?
> * **Update generated TS api**
> * **Fix a bug where local environment used wrong OAuth2 redirect URL**
> * Added the "MoleTable"
> * Added tests for the "MoleTable"
> * **Enable yarn tests in the CI**
> * **Removed App.test.tsx as a lot of mocking is required to test the AuthProvider**

I put all the changes that doesn't belong in this PR in bold. You can open separate PRs for these and open a new PR to close #23. If you need help to reorganise the code into separate branches, we can set up a meeting to look at it together.

PR #40, opened 2022-09-26
fix: update generated TS api

soofstad:
- Get new endpoints  and models into generated API


mariuswinger (approved):
something weird with pre-commit. Rest looks good.

web/src/api/MercuryAPI.tsx
```diff
@@ -1,9 +1,10 @@
-import { Configuration, TodosApi } from './generated'
+import { ComponentApi, Configuration } from './generated'
 
-export class MercuryAPI extends TodosApi {
+// TODO: Combine all generated api's into MercuryAPI
+export class MercuryAPI extends ComponentApi {
   constructor(token: string) {
     const configuration = new Configuration({
-      basePath: 'http://localhost',
+      basePath: '',
```
olavis, line 7: Maybe use a constant for basePath?

web/src/api/MercuryAPI.tsx
```diff
@@ -1,9 +1,10 @@
-import { Configuration, TodosApi } from './generated'
+import { ComponentApi, Configuration } from './generated'
 
-export class MercuryAPI extends TodosApi {
+// TODO: Combine all generated api's into MercuryAPI
```
olavis, line 3: Do we have an issue for this? I don't understand.

web/src/api/MercuryAPI.tsx
soofstad: There is no value. `http://localhost` has been removed.

web/src/api/MercuryAPI.tsx
soofstad: Because of the tags we use on the backend, the generated API is split into 4 different ones.
```
// from web/src/api/generated/api.ts
export * from './api/component-api';
export * from './api/health-check-api';
export * from './api/multiflash-api';
export * from './api/whoami-api';
```

This will lead to duplicated api wrappers like "MercuryAPI" if not combined in some way

PR #41, opened 2022-09-26
fix: response/input is now camelcase

mariuswinger:

This pull request is needed because we want request and response models to be camelcase.


api/src/tests/integration/features/multiflash/test_multiflash_feature.py
```diff
@@ -15,8 +15,9 @@
 def test_compute_multiflash_feature(test_app, multiflash_input: dict, multiflash_expected_output: dict):
     response = test_app.post("/api/v1/multiflash", json=multiflash_input)
     results = response.json()
-    computed_phase_values = results["phase_values"]
-    computed_component_fractions = results["component_fractions"]
+    print(results)
```
olavis, line 18: Did you forget to remove print?

mariuswinger: :)))))

olavis (dismissed):

olavis (approved):

api/src/features/component/component_feature.py
```diff
@@ -5,11 +5,12 @@
...
 router = APIRouter(tags=["component"])
 
 
 @create_response(JSONResponse)
-@router.get("/components", operation_id="get_components", response_model=Dict[int, Dict[str, str]])
-async def components() -> Dict[int, Dict[str, str]]:
-    return COMPONENT_IDS
+@router.get("/components", operation_id="get_components", response_model=Dict[str, Dict[str, str]])
+async def components() -> Dict[str, Dict[str, str]]:
+    return component_use_case(COMPONENT_IDS)
```
soofstad, line 16: For the TS generated types, a BaseModel class for the response is needed.
`components() -> ComponentIds:
    return ComponentsIds(**COMPONENT_IDS)

Would also make this use_case unnecessary

soofstad (changes requested):
When making changes to the api affecting the TS generated code.
It would be nice if we were to regenerate it, and see if the result was as expected.

e.g. This this camelCase change

api/src/tests/integration/features/component/test_component_feature.py
```diff
@@ -1,14 +1,15 @@
...
     response = test_app.get("/api/v1/components")
     results = response.json()
-    # json turns ints to str, need to convert back for testing:
-    # using sets as sets dont care about ordering
-    assert {key for key in results.keys()} == set(COMPONENT_IDS.keys())
-    for key, value in results.items():
-        assert value == COMPONENT_IDS[key]
+    components = results["components"]
+    # using sets as sets do not care about ordering
+    assert {key for key in components.keys()} == set(COMPONENT_IDS.keys())
```
soofstad, line 12:
```suggestion
    assert set(components.keys()) == set(COMPONENT_IDS.keys())
```
This will work right?

soofstad (approved):

api/src/tests/integration/features/component/test_component_feature.py
mariuswinger: Yes.

api/src/features/component/component_feature.py
mariuswinger: fixed

PR #16, opened 2022-08-19
feat: header, user info, and about

soofstad:
Added a EDS styled header, along with a clickable about and user info
icon

PR #45, opened 2022-09-26
fix: broken auth on local envinronment

soofstad:

- Should be able to run the web app with and without authentication in local environment 

- fix missing redirect url for login
- use AuthProvider state to check if loginInProgres, and conditionally don't render any components
- Update Header with more sensable user info when none available


web/src/App.tsx
```diff
@@ -2,7 +2,7 @@ import { AuthProvider } from 'react-oauth2-code-pkce'
 import { authConfig } from './authConfig'
 import { MainPage } from './pages/Main'
 
-const AUTH_DISABLED = process.env.REACT_APP_AUTH !== '1'
+export const AUTH_DISABLED = process.env.REACT_APP_AUTH !== '1'
```
olavis, line 5: I'm not sure if App.tsx is the best owner of this constant, now that it is used globally. Maybe we could store constants in a separate file, so we'll know where to find them?

web/.eslintignore
```diff
@@ -3,4 +3,3 @@ web/src/api/generated
 
 build
 coverage
-web/src/components/TodoApp.tsx
```
olavis, line 6: Good that you removed this. Next time, maybe we should do it in a separate commit to keep order in the git history. I don't see why this change belongs under the title `fix: broken auth under local environment`

web/src/components/common/Header.tsx
```diff
@@ -3,6 +3,7 @@ import { TopBar, Icon, Popover, Button } from '@equinor/eds-core-react'
 import { account_circle, grid_on, info_circle } from '@equinor/eds-icons'
 import { useContext, useRef, useState } from 'react'
 import { AuthContext } from 'react-oauth2-code-pkce'
+import { AUTH_DISABLED } from '../../App'
```
olavis, line 6: Again, not sure if this constant should be exported from App.tsx. I find it hard to agree that Header is a common component now that it is dependent on another (non-common) component.

olavis (changes requested):

web/yarn.lock
```diff
@@ -10516,12 +10516,12 @@ __metadata:
...
-  version: 1.4.0
-  resolution: "react-oauth2-code-pkce@npm:1.4.0"
+"react-oauth2-code-pkce@npm:^1.7.1":
+  version: 1.7.1
+  resolution: "react-oauth2-code-pkce@npm:1.7.1"
   peerDependencies:
-    react: ^17.0.2
-  checksum: 3cb2d7126926c8538f101e03b5cee853b555692b41daef751e1509a99c582400a6c352099dc2fe502f7521520c0bad47eba31dd6bf57b114061380969f55144b
+    react: ">=16.8.0"
+  checksum: f841c1ce49b37e90b35919dc8ff207ccb10bab0d08c6ac9040ff6874b4c53207eaf41201db78c191247deaaa4ee9edc5470df200c693cd8aa433355e616cc87b
```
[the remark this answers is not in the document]
soofstad: No, it will fix that as well. But that was a side effect. The reason for the upgrade was for the "loginInProgress" feature

web/src/pages/Main.tsx
```diff
@@ -21,6 +23,8 @@ export const MainPage = (): JSX.Element => {
             related to the app.
           </p>
         </div>
+      ) : (
+        <div>This is my app</div>
```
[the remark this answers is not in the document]
soofstad: This has not been changed in this PR. Ether way it's a placeholder that will be replaced soon.

web/src/components/common/Header.tsx
soofstad: Agreed. Header is not "common". It's only used one place.
But out of scope for this PR/issue

web/src/pages/Main.tsx
```diff
@@ -1,15 +1,17 @@
 import { Header } from '../components/common/Header'
 import { useContext } from 'react'
 import { AuthContext } from 'react-oauth2-code-pkce'
+import { AUTH_DISABLED } from '../App'
 
 export const MainPage = (): JSX.Element => {
-  const { token } = useContext(AuthContext)
+  const { token, loginInProgress } = useContext(AuthContext)
+
+  if (loginInProgress) return <></>
```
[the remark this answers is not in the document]
soofstad: No-can-do. We need at least one level of components between the AuthProvider the the "useContext".
Could of course add another component between App and Main. But as long as we only have one page. This seems like the best solution.

web/src/pages/Main.tsx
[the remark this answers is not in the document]
olavis: Well, the line _has_ changed in this PR. Maybe I should remind you that it was addressed in the PR where it was introduced as well: https://github.com/equinor/mercury/pull/16 The conversation was resolved, but it doesn't seem like anything was done? And no reply? Am I supposed to just shut my mouth if I recognise issues I earlier have pointed out that has been ignored?

web/src/components/common/Header.tsx
olavis: The change was introduced in this PR, so if it's out of scope, well then changing the `AUTH_DISABLED` constant to a global is out of scope as well. The Header wasn't dependent on App.tsx until now.

web/src/App.tsx
soofstad: Fixed

web/src/pages/Main.tsx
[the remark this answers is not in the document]
olavis: I'll see if I can provide an example

web/src/pages/Main.tsx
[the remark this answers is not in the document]
soofstad: No. Oversight in #16.

web/src/constants.ts
```diff
@@ -0,0 +1 @@
+export const AUTH_DISABLED = process.env.REACT_APP_AUTH !== '1'
```
olavis, line 1: 👍

mariuswinger (approved):

PR #46, opened 2022-09-26
feature: MoleTable and DynamicTable

### LINKED ISSUE #23: Implement results table

soofstad:

Users should be able to see the mole results from the calculations

- Added the "MoleTable"
- Added tests for the "MoleTable"

closes #23

### DIFF
--- a/web/src/components/common/DynamicTable.tsx
+++ b/web/src/components/common/DynamicTable.tsx
@@ -1,0 +1,43 @@
+import { EdsProvider, Table } from '@equinor/eds-core-react'
+
+export type TDynamicTableInput = {
+  // Keys will be used as headers. The array for each key will be the data in the column
+  [key: string]: Array<string>
+}
+export const DynamicTable = ({
+  input,
+}: {
+  input: TDynamicTableInput
+}): JSX.Element => {
+  function createTableRows() {
+    const rows: Array<JSX.Element> = []
+
+    input[Object.keys(input)[0]].forEach((value, index) => {
+      const cells: Array<JSX.Element> = []
+      Object.keys(input).forEach((key) => {
+        cells.push(
+          <Table.Cell data-testid={`${key}-${index}`} key={`${key}-${index}`}>
+            {input[key][index]}
+          </Table.Cell>
+        )
+      })
+      rows.push(<Table.Row key={index}>{cells}</Table.Row>)
+    })
+    return rows
+  }
+
+  return (
+    <EdsProvider density={'compact'}>
+      <Table>
+        <Table.Head>
+          <Table.Row>
+            {Object.keys(input).map((header) => (
+              <Table.Cell key={header}>{header}</Table.Cell>
+            ))}
+          </Table.Row>
+        </Table.Head>
+        <Table.Body>{createTableRows()}</Table.Body>
+      </Table>
+    </EdsProvider>
+  )
+}
--- a/web/src/components/feature/MoleTable.test.tsx
+++ b/web/src/components/feature/MoleTable.test.tsx
@@ -1,0 +1,76 @@
+import { render, screen } from '@testing-library/react'
+import { MoleTable } from './MoleTable'
+import '@testing-library/jest-dom/extend-expect'
+import { ComponentResponse, MultiflashResponse } from '../../api/generated'
+
+const testComponents: ComponentResponse = {
+  components: {
+    '1': { chemicalFormula: 'CO2', altName: 'Carbondioxide' },
+    '2': { chemicalFormula: 'N2', altName: 'Nitrogen' },
+    '3': { chemicalFormula: 'H2O', altName: 'Water' },
+    '4': { chemicalFormula: 'H2S', altName: 'Hydrogensulfide' },
+    '5': { chemicalFormula: 'Hg', altName: 'Mercury' },
+    '101': { chemicalFormula: 'CH4', altName: 'Methane' },
+    '201': { chemicalFormula: 'C2H6', altName: 'Ethane' },
+    '301': { chemicalFormula: 'nC3', altName: 'Propane' },
+    '401': { chemicalFormula: 'iC4', altName: 'i-Butane' },
+    '402': { chemicalFormula: 'nC4', altName: 'n-Butane' },
+    '501': { chemicalFormula: '22-dm-C3', altName: '2-2-dimethyl-propane' },
+    '503': { chemicalFormula: 'iC5', altName: 'i-Pentane' },
+    '504': { chemicalFormula: 'nC5', altName: 'n-Pentane' },
+    '502': { chemicalFormula: 'cy-C5', altName: 'Cy-C5' },
+    '601': {
+      chemicalFormula: '22-dm-C4',
+      altName: '2-2-dimethyl-butane(neohexane)',
+    },
+    '602': { chemicalFormula: '23-dm-C4', altName: '2-3-dimethyl-butane' },
+    '603': { chemicalFormula: '2-m-C5', altName: '2-methyl-pentane' },
+    '604': { chemicalFormula: '3-m-C5', altName: '3-methyl-pentane' },
+    '605': { chemicalFormula: 'nC6', altName: 'n-Hexane' },
+    '606': { chemicalFormula: 'cy-C6', altName: 'Cy-hexane' },
+    '608': { chemicalFormula: 'benzene', altName: 'Benzene' },
+    '701': { chemicalFormula: 'nC7', altName: 'n-heptane' },
+    '707': { chemicalFormula: 'cy-C7', altName: 'Cy-heptane' },
+    '710': { chemicalFormula: 'toluene', altName: 'Toluene' },
+    '801': { chemicalFormula: 'nC8', altName: 'n-octane' },
+    '806': { chemicalFormula: 'cy-C8', altName: 'Cy-octane' },
+    '809': { chemicalFormula: 'm-xylene', altName: 'M-xylene' },
+    '901': { chemicalFormula: 'nC9', altName: 'n-nonane' },
+    '1016': { chemicalFormula: 'nC10', altName: 'n-decane' },
+  },
+}
+
+const testMultiflashResult: MultiflashResponse = {
+  phaseValues: {
+    Vapor: { percentage: 0.9989992588, mercury: 104.82781237 },
+    Mercury: { percentage: 0.0010007412, mercury: 1000000000.0 },
+  },
+  componentFractions: {
+    2: [0.01693, 0.000000000000000000000003431],
+    1: [0.05399, 0.000000000000000000000006807],
+    101: [0.8563, 0.0000000000000000000001379],
+    201: [0.04577, 0.00000000000000000000000481],
+    301: [0.01763, 0.000000000000000000000001301],
+    401: [0.002424, 0.0000000000000000000000001389],
+    402: [0.004378, 0.0000000000000000000000002324],
+    503: [0.001212, 0.00000000000000000000000005006],
+    504: [0.001102, 0.00000000000000000000000004238],
+    608: [0.0001473, 0.000000000000000000000000005109],
+    710: [0.00008415, 0.000000000000000000000000002175],
+    809: [0.00002304, 0.000000000000000000000000000433],
+    5: [0.00000001236, 1.0],
+  },
+}
+
+test('renders without crashing and displaying correct values in table', async () => {
+  render(
+    <MoleTable
+      components={testComponents}
+      multiFlashResponse={testMultiflashResult}
+    />
+  )
+  // @ts-ignore because not able to get eslint to discover these types
+  expect(screen.getByTestId('Vapor-0')).toHaveTextContent('0.05399')
+  // @ts-ignore because not able to get eslint to discover these types
+  expect(screen.getByTestId('Mercury-7')).toHaveTextContent('2.324e-25')
+})
--- a/web/src/components/feature/MoleTable.tsx
+++ b/web/src/components/feature/MoleTable.tsx
@@ -1,0 +1,45 @@
+import { DynamicTable, TDynamicTableInput } from '../common/DynamicTable'
+import { ComponentResponse, MultiflashResponse } from '../../api/generated'
+
+function transformRawResponse(
+  multiFlashResponse: MultiflashResponse,
+  componentResponse: ComponentResponse
+): TDynamicTableInput {
+  const tableInput: TDynamicTableInput = { Components: [] }
+  const phasesInResult: string[] = Object.keys(multiFlashResponse.phaseValues)
+
+  // Add phases "headers" to the tableInput
+  phasesInResult.forEach((phase) => {
+    tableInput[phase] = []
+  })
+
+  // Push data to each phase column
+  phasesInResult.forEach((phase, index) => {
+    Object.values(multiFlashResponse.componentFractions).forEach(
+      (fractions: number[]) => {
+        tableInput[phase].push(fractions[index].toString())
+      }
+    )
+  })
+
+  // Create the components column
+  Object.keys(multiFlashResponse.componentFractions).forEach((compId) => {
+    tableInput.Components.push(componentResponse.components[compId].altName)
+  })
+
+  return tableInput
+}
+
+// TODO: Get type from generated API
+export const MoleTable = (props: {
+  multiFlashResponse: MultiflashResponse
+  components: ComponentResponse
+}) => {
+  const { multiFlashResponse, components } = props
+
+  return (
+    <DynamicTable
+      input={transformRawResponse(multiFlashResponse, components)}
+    />
+  )
+}
--- a/web/src/pages/Main.tsx
+++ b/web/src/pages/Main.tsx
@@ -1,24 +1,66 @@
 import { Header } from '../components/common/Header'
-import { useContext, useState } from 'react'
+import { useEffect, useContext, useState } from 'react'
+import { MoleTable } from '../components/feature/MoleTable'
 import { AuthContext } from 'react-oauth2-code-pkce'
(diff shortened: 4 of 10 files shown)


web/src/components/feature/MoleTable.tsx
```diff
@@ -0,0 +1,51 @@
+import { DynamicTable, TDynamicTableInput } from '../common/DynamicTable'
+import { TComponentResponse } from '../../pages/Main'
+
+// TODO: Update with type from generated api
+type TMultiFlashResponse = {
+  component_fractions: { [key: number]: number[] }
+  phase_values: { [key: string]: { percentage: number; mercury: number } }
```
olavis, line 7: We don't do snake case in typescript 📘

web/src/components/feature/MoleTable.tsx
```diff
@@ -0,0 +1,51 @@
+import { DynamicTable, TDynamicTableInput } from '../common/DynamicTable'
+import { TComponentResponse } from '../../pages/Main'
```
olavis, line 2: This is not exported from Main as far as I know?

olavis:
I'm not able to build the app atm, but I'll finish this review as soon as I get it up and running again.

web/src/components/feature/MoleTable.tsx
soofstad: this structure represents the API response as it is

web/src/components/feature/MoleTable.tsx
soofstad: Fixed. Changes made to Main.tsx was missing. Added now

web/src/components/feature/MoleTable.tsx
mariuswinger: Not anymore!

web/src/pages/Main.tsx
```diff
@@ -1,14 +1,65 @@
...
+}
 
 export const MainPage = (): JSX.Element => {
-  const { token } = useContext(AuthContext)
+  const [components, setComponents] = useState<TComponentResponse>()
+  const [loading, setLoading] = useState<boolean>(false)
+
+  // Fetch list of components name once on page load
+  useEffect(() => {
+    const mercuryApi = new MercuryAPI('')
```
olavis, line 18: This violates the Dependency Inversion Principle. Main will now create a new instance of the API on every render.. 
I suggest we make a custom hook for this instead.

web/src/pages/Main.tsx
```diff
@@ -1,14 +1,65 @@
...
+
+// TODO: Get type from generated API
+export type TComponentResponse = {
+  [key: string]: { [key: string]: string }
+}
 
 export const MainPage = (): JSX.Element => {
-  const { token } = useContext(AuthContext)
+  const [components, setComponents] = useState<TComponentResponse>()
+  const [loading, setLoading] = useState<boolean>(false)
```
olavis, line 14: Can we use naming conventions for booleans? `isLoading`?

web/src/pages/Main.tsx
```diff
@@ -1,14 +1,65 @@
...
+      .getComponents()
+      .then((response: AxiosResponse<TComponentResponse>) =>
+        setComponents(response.data)
+      )
+      .finally(() => setLoading(false))
+  }, [])
+
+  // TODO: Remove with actual calculation results
+  const multiflashResult = {
+    phase_values: {
```
olavis, line 29: Snake_case 📘

web/src/pages/Main.tsx
```diff
@@ -1,14 +1,65 @@
...
+      Vapor: { percentage: 0.9989992588, mercury: 104.82781237 },
+      Mercury: { percentage: 0.0010007412, mercury: 1000000000.0 },
+    },
+    component_fractions: {
```
olavis, line 33: Snake_case 📘

web/src/pages/Main.tsx
```diff
@@ -1,14 +1,65 @@
...
+        setComponents(response.data)
+      )
+      .finally(() => setLoading(false))
+  }, [])
+
+  // TODO: Remove with actual calculation results
+  const multiflashResult = {
+    phase_values: {
+      Vapor: { percentage: 0.9989992588, mercury: 104.82781237 },
+      Mercury: { percentage: 0.0010007412, mercury: 1000000000.0 },
```
olavis, line 31: `Vapor` and `Mercury` should be in camelCase

web/src/pages/Main.tsx
```diff
@@ -1,14 +1,65 @@
...
+
+  // TODO: Better error handling and message
+  if (!components) return <div>Failed to fetch list of components</div>
+
   return (
     <div>
       <Header />
-      {token ? (
-        <div>This is my app</div>
+      {true ? (
```
olavis, line 58: Should avoid constant expressions. Theres no need to check if `true` is `true`

web/src/pages/Main.tsx
```diff
@@ -1,14 +1,65 @@
...
+      710: [0.00008415, 0.000000000000000000000000002175],
+      809: [0.00002304, 0.000000000000000000000000000433],
+      5: [0.00000001236, 1.0],
+    },
+  }
+
+  if (loading) return <>Loading...</>
+
+  // TODO: Better error handling and message
+  if (!components) return <div>Failed to fetch list of components</div>
```
olavis, line 53: This will be true until components are fetched, but that doesn't necessarily mean that fetching failed.

web/src/pages/Main.tsx
```diff
@@ -1,14 +1,65 @@
...
+      503: [0.001212, 0.00000000000000000000000005006],
+      504: [0.001102, 0.00000000000000000000000004238],
+      608: [0.0001473, 0.000000000000000000000000005109],
+      710: [0.00008415, 0.000000000000000000000000002175],
+      809: [0.00002304, 0.000000000000000000000000000433],
+      5: [0.00000001236, 1.0],
+    },
+  }
+
+  if (loading) return <>Loading...</>
```
olavis, line 50: I prefer this `{isLoading && <>Loading...</>}` and trying to avoid having multiple returns in a component because it's easier to control what the component should return.

web/src/pages/Main.tsx
```diff
@@ -1,14 +1,65 @@
...
+
+  // TODO: Better error handling and message
+  if (!components) return <div>Failed to fetch list of components</div>
+
   return (
     <div>
```
olavis, line 56: This section could be replaced with:
```
<>
   <Header />
   <MoleTable
     multiFlashResponse={multiflashResult}
     components={components}
   />
</>
```
IMO that's more readable.

web/src/pages/Main.tsx
```diff
@@ -1,14 +1,65 @@
...
 export const MainPage = (): JSX.Element => {
-  const { token } = useContext(AuthContext)
+  const [components, setComponents] = useState<TComponentResponse>()
+  const [loading, setLoading] = useState<boolean>(false)
+
+  // Fetch list of components name once on page load
+  useEffect(() => {
+    const mercuryApi = new MercuryAPI('')
+    mercuryApi
+      .getComponents()
```
olavis, line 20: Does this exist?

web/src/components/common/DynamicTable.tsx
```diff
@@ -0,0 +1,43 @@
+import { EdsProvider, Table } from '@equinor/eds-core-react'
+
+export type TDynamicTableInput = {
+  // Keys will be used as headers. The array for each key will be the data in the column
+  [key: string]: Array<string>
+}
+export const DynamicTable = (props: {
+  input: TDynamicTableInput
+}): JSX.Element => {
+  const { input } = props
```
olavis, line 10: is it really necessary to add a const when we have just one param? I think this is more readable: 
`DynamicTable = ({ input }: { input: TDynamicTableInput })`

web/src/components/common/DynamicTable.tsx
```diff
@@ -0,0 +1,43 @@
...
+
+  function createTableBody() {
+    const rows: Array<JSX.Element> = []
+
+    input[Object.keys(input)[0]].forEach((value, index) => {
```
olavis, line 15: `value` is unused

web/src/components/common/DynamicTable.tsx
```diff
@@ -0,0 +1,43 @@
...
+export type TDynamicTableInput = {
+  // Keys will be used as headers. The array for each key will be the data in the column
+  [key: string]: Array<string>
+}
+export const DynamicTable = (props: {
+  input: TDynamicTableInput
+}): JSX.Element => {
+  const { input } = props
+
+  function createTableBody() {
```
olavis, line 12: The function doesn't really create a table body tho.. maybe we should find a more suitable name for it

web/src/components/feature/MoleTable.test.tsx
```diff
@@ -0,0 +1,73 @@
+import { render, screen } from '@testing-library/react'
+import { MoleTable } from './MoleTable'
+import '@testing-library/jest-dom/extend-expect'
+
```
olavis, line 4: We should be consistent with casing. This file has a little bit of everything.

web/src/components/feature/MoleTable.test.tsx
```diff
@@ -0,0 +1,73 @@
...
+    503: [0.001212, 0.00000000000000000000000005006],
+    504: [0.001102, 0.00000000000000000000000004238],
+    608: [0.0001473, 0.000000000000000000000000005109],
+    710: [0.00008415, 0.000000000000000000000000002175],
+    809: [0.00002304, 0.000000000000000000000000000433],
+    5: [0.00000001236, 1.0],
+  },
+}
+
+test('renders without crashing and displaying correct values in table', async () => {
```
olavis, line 62: A test should only test _one_ thing. If this fails, is it because it doesn't render or because the table doesn't have correct values?

web/src/components/feature/MoleTable.test.tsx
```diff
@@ -0,0 +1,73 @@
...
+  render(
+    <MoleTable
+      components={testComponents}
+      multiFlashResponse={testMultiflashResult}
+    />
+  )
+  // @ts-ignore because not able to get eslint to discover these types
```
olavis, line 69: IMO this is not a valid reason to ignore. And it's nothing wrong with eslint.

web/src/components/feature/MoleTable.tsx
soofstad: Right. This should be updated when the code is updated to do an actual API call

web/src/components/feature/MoleTable.test.tsx
```diff
@@ -0,0 +1,73 @@
...
+  expect(screen.getByTestId('Vapor-0')).toHaveTextContent('0.05399')
+  // @ts-ignore because not able to get eslint to discover these types
```
olavis, line 71: Not valid/correct reason. The property doesn't exist.

web/src/components/feature/MoleTable.tsx
```diff
@@ -0,0 +1,51 @@
...
+
+// TODO: Update with type from generated api
+type TMultiFlashResponse = {
+  component_fractions: { [key: number]: number[] }
+  phase_values: { [key: string]: { percentage: number; mercury: number } }
+}
+
+function transformRawResponse(
```
olavis, line 10: This function is doing a lot

olavis (changes requested):
See comments

web/src/pages/Main.tsx
soofstad: Bad change from splitting the original PR. Fixed

web/src/pages/Main.tsx
soofstad: done

web/src/pages/Main.tsx
soofstad: the temp mock data represent the values returned from the API

web/src/pages/Main.tsx
soofstad: the temp mock data represent the values returned from the API

olavis:
This PR is getting old/big. Should we close it and try not to bite off more than one can chew? I see that it can be divided into smaller tasks, implementing a custom hook to fetch components for instance. It's already complex enough to go through multiple rounds of reviews.

web/src/pages/Main.tsx
soofstad: the temp mock data represent the values returned from the API

web/src/pages/Main.tsx
soofstad: fixed

web/src/pages/Main.tsx
soofstad: Right. That why it says `//TODO`

web/src/pages/Main.tsx
soofstad: Disagree. Having multiple nested conditional inline if-else renders in a single return statement will be less readable/maintanable.

web/src/pages/Main.tsx
soofstad: yes

web/src/components/common/DynamicTable.tsx
soofstad: Agreed. Fixed

web/src/components/common/DynamicTable.tsx
soofstad: yes. But needed to access the ìndex` value

web/src/components/common/DynamicTable.tsx
soofstad: Agreed. Renamed

web/src/pages/Main.tsx
olavis: It's still a problem though, if you create an issue on it and link it I'm fine with that. Just trying not to overlook potential issues.

web/src/pages/Main.tsx
olavis: Okay, we'll go with that, then. As long as the file isn't too big and the returns are easy to find, I guess it won't be a problem.

web/src/components/feature/MoleTable.test.tsx
soofstad: Which parts are not consistent?

web/src/components/feature/MoleTable.test.tsx
soofstad: one implies the other. Changed to more obvious naming

web/src/components/feature/MoleTable.test.tsx
soofstad: It exists. Or the the test would have failed.
Any suggestions are welcome

web/src/components/feature/MoleTable.tsx
soofstad: yes

web/src/components/feature/MoleTable.test.tsx
soofstad: issue created

soofstad: #57

web/src/components/feature/MoleTable.tsx
soofstad: considered. Keeping it

web/src/pages/Main.tsx
soofstad: This will be resolved when implementing the "run calculation" issue

web/src/components/feature/MoleTable.tsx
```diff
@@ -0,0 +1,45 @@
+import { DynamicTable, TDynamicTableInput } from '../common/DynamicTable'
+import { ComponentResponse, MultiflashResponse } from '../../api/generated'
+
+function transformRawResponse(
+  multiFlashResponse: MultiflashResponse,
+  componentResponse: ComponentResponse
+): TDynamicTableInput {
+  const tableInput: TDynamicTableInput = { Components: [] }
```
olavis, line 8: components in lowercase

web/src/pages/Main.tsx
olavis: I think it got lost in the rebase?

web/src/pages/Main.tsx
olavis: You can move mercuryApi outside the useEffect, to prevent unnecessary initialisations. You can also wrap the fetch in an expression to check if components has a state, to prevent re-fetching.

web/src/components/feature/MoleTable.tsx
soofstad: It's used here as the actual header of the table.
I'd rather not do any extra string manipulation that can take away flexibility of the header value.

soofstad: Could do `['Components']` to signal that it is meant to be a friendly string value. But eslint does not like that (Unnecessarily computed property ['Components'] found  no-useless-computed-key).

web/src/pages/Main.tsx
soofstad: Yes. fixed :)

web/src/pages/Main.tsx
soofstad: Moved api init to one level up.
The useEffect has an empty dependency array, and will only run once, when components has no state.

web/src/components/common/DynamicTable.tsx
olavis: I think you forgot?

soofstad: Damn rebase....

web/src/components/common/DynamicTable.tsx
olavis: Not fixed?

soofstad: now

mariuswinger (approved):

PR #52, opened 2022-09-27
Fix/ci tests

mariuswinger:

This pull request is needed because ci was unstable.

Environmental variables for ifort is set by shell script provided by intel oneapi.
Add githash variable before clone to ensure docker caching works when git hash is updated.
Pre-commit caching is disabled (very unstable).


soofstad (approved):
:+1:

soofstad (approved):

PR #56, opened 2022-09-28
ci: update githash for libhg

mariuswinger:
Updates the git hash after merging in libhg repo.


soofstad (approved):
I assume that is the correct hash

mariuswinger:
> I assume that is the correct hash

Screenshot 2022-09-29 at 12 07 26

Issue #62, opened 2022-10-06
Add table to display mercury distribution results


mariuswinger:

mariuswinger:
Closed by #64

PR #64, opened 2022-10-07
feat: add PhaseTable

mariuswinger:

Want to have a table to show results of phases.

Adds that table.

#62

### DIFF
--- a/web/src/components/feature/PhaseTable.tsx
+++ b/web/src/components/feature/PhaseTable.tsx
@@ -1,0 +1,46 @@
+import { DynamicTable, TDynamicTableInput } from '../common/DynamicTable'
+import { MultiflashResponse } from '../../api/generated'
+
+function transformRawResponse(
+  multiFlashResponse: MultiflashResponse,
+  cubicFeedFlow: number
+): TDynamicTableInput {
+  const tableInput: TDynamicTableInput = {
+    Phases: [],
+    [`Mass Concentration`]: [],
+    [`Mole Concentration`]: [],
+    [`Mercury Flow (g/d)`]: [],
+  }
+  const phasesInResult: string[] = Object.keys(multiFlashResponse.phaseValues)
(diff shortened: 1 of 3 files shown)


web/src/components/feature/PhaseTable.tsx
```diff
@@ -0,0 +1,45 @@
...
+    [`Mercury Flow (g/d)`]: [],
+  }
+  const phasesInResult: string[] = Object.keys(multiFlashResponse.phaseValues)
+  phasesInResult.forEach((phase, index) => {
+    tableInput['Phases'].push(phase)
+    tableInput['Mass Concentration'].push(
+      multiFlashResponse.phaseValues[phase]['mercury'].toString()
+    )
+    tableInput['Mole Concentration'].push(
+      multiFlashResponse.componentFractions['5'][index].toString()
```
soofstad, line 21: I assume the '5' is the id for mercury.
But coming back after 2 years, that would not be obvious. A small comment would help.

web/src/components/feature/PhaseTable.tsx
```diff
@@ -0,0 +1,45 @@
...
+    )
+    // TODO: add support for feed in kg/d
+    const hgPhaseFlow =
+      cubicFeedFlow *
+      42.29256 *
```
soofstad, line 26: Do we have an explanation for these constants?
Can that be added as a comment?

soofstad (changes requested):
Good stuff!

Would like to see a test of the PhaseTable component.
Can pretty much just copy the MoleTable.test.tsx

web/src/components/feature/PhaseTable.tsx
mariuswinger: No explanation. I'm guessing it is interpolated from empirical data.

soofstad (approved):
Good. If you rebase first, you will get a type error in the MockData in the test

PR #66, opened 2022-10-07
Fix/add total mol weight to response

mariuswinger:
Need total molecular weight to convert feed flow between kg/d and sm3/d.

Changes the MultiflashResponse to contain molecular weight as a property.


api/src/common/utils/tuples.py
```diff
@@ -52,6 +52,7 @@ class MultiflashResult(NamedTuple):
 
     phase_values: Dict[PhaseLabels, PhaseValues]
     component_fractions: Dict[str, ComponentFractions]
+    feed_molecular_weight: float
```
soofstad, line 55: Add this to the class comments?

soofstad (approved):

api/src/common/utils/tuples.py
mariuswinger: will do

PR #67, opened 2022-10-10
feat: add feed unit input and switch

### LINKED ISSUE #73: Add feed flow switch component

mariuswinger:

In order to compute the PhaseTable component, it is necessary to have a number for the feed flow (this is input, but not input to the Fortran code). The feed flow can be given as in Sm3/d or kg/d depending on which information the user has available, therefore there is a switch to change units.

Add a text field and a switch.

Closes #73

### DIFF
--- a/web/src/components/feature/FeedFlowInput.tsx
+++ b/web/src/components/feature/FeedFlowInput.tsx
@@ -1,0 +1,38 @@
+import { Switch, TextField } from '@equinor/eds-core-react'
+import styled from 'styled-components'
+import { TFeedUnit } from './PhaseTable'
+
+const FlexContainer = styled.div`
+  display: flex;
+  gap: 16px;
+`
+
+export const FeedFlowInput = (props: {
+  feedFlow: number
+  feedUnit: TFeedUnit
+  setFeedFlow: (value: number) => void
+  setFeedUnit: (unit: TFeedUnit) => void
+}) => {
+  return (
+    <FlexContainer>
+      <TextField
+        id="feed-flow-input"
+        value={props.feedFlow.toString()}
+        label="Feed Flow"
+        unit={props.feedUnit}
+        type="number"
+        onChange={(event: React.ChangeEvent<HTMLInputElement>) =>
+          props.setFeedFlow(Number(event.target.value))
+        }
+      />
+      <Switch
+        label="Feed unit"
+        size="small"
+        checked={props.feedUnit === 'kg/d'}
+        onChange={(event: React.ChangeEvent<HTMLInputElement>) =>
+          props.setFeedUnit(event.target.checked ? 'kg/d' : 'Sm3/d')
+        }
+      />
+    </FlexContainer>
+  )
+}
--- a/web/src/components/feature/PhaseTable.tsx
+++ b/web/src/components/feature/PhaseTable.tsx
@@ -1,11 +1,24 @@
 import { DynamicTable } from '../common/DynamicTable'
 import { MultiflashResponse } from '../../api/generated'
+import styled from 'styled-components'
+import { FeedFlowInput } from './FeedFlowInput'
+import { useState } from 'react'
+
+export type TFeedUnit = 'kg/d' | 'Sm3/d'
+
+const PhaseTableContainer = styled.div`
+  display: flex;
+  flex-direction: column;
+  row-gap: 4px;
+  max-width: 500px;
+`
 
 function getRows(
   multiFlashResponse: MultiflashResponse,
-  cubicFeedFlow: number
+  feedFlow: number
 ): string[][] {
-  const phPhaseFlowFactor = cubicFeedFlow * 42.29256 * 200.59
+  // TODO: Assumes feed flow unit is Sm3/d
+  const phPhaseFlowFactor = feedFlow * 42.29256 * 200.59
   const mercury = multiFlashResponse.componentFractions['5']
   return Object.entries(multiFlashResponse.phaseValues).map(
     ([phase, values], index) => [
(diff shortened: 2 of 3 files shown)


web/src/components/feature/CalculateFluid.tsx
```diff
@@ -67,6 +74,24 @@ export const CalculateFluid = ({
...
+          id="feed-flow-input"
+          value={feedFlow.toString()}
+          label="Feed Flow"
+          unit={feedUnit}
+          type="number"
+          onChange={(event: React.ChangeEvent<HTMLInputElement>) =>
+            setFeedFlow(Number(event.target.value))
+          }
+        />
+        <Switch
```
soofstad, line 87: According to the figma sketches. This component should be placed with the "PhaseTable".
I see no reason to change that.

soofstad (changes requested):

web/src/components/feature/CalculateFluid.tsx
mariuswinger: Did we conclude to await meeting on this one?

soofstad: Concluded we do as in the sketches, and ask for feedback before moving the component somewhere else.

mariuswinger: Will fix

mariuswinger:
Screenshot 2022-10-11 at 15 20 33

web/src/components/feature/FeedFlowInput.tsx
```diff
@@ -0,0 +1,33 @@
...
+import styled from 'styled-components'
+
+const FlexContainer = styled.div`
+  display: flex;
+  gap: 16px;
+`
+
+export const FeedFlowInput = (props: {
+  feedFlow: number
+  feedUnit: string
```
ingeridhellen, line 11:
```suggestion
  feedUnit: 'kg/d' | 'Sm3/d'
```

web/src/components/feature/PhaseTable.tsx
```diff
@@ -19,18 +30,36 @@ function getRows(
...
-        'Mole Concentration',
-        'Mercury Flow (g/d)',
-      ]}
-      rows={getRows(multiFlashResponse, cubicFeedFlow)}
-    />
+    <PhaseTableContainer>
+      <FeedFlowInput
+        feedFlow={feedFlow}
+        feedUnit={feedUnit}
+        onChangeFeedFlow={onChangeFeedFlow}
```
soofstad, line 51: This is a matter if taste. But IMO, a simple "setState" should be done inline. Results in less jumping around when reading the code.
`onChangeFeedFlow={(...)=>setState(Number(event.target.value))}`

web/src/components/feature/FeedFlowInput.tsx
```diff
@@ -0,0 +1,33 @@
...
+}) => {
+  return (
+    <FlexContainer>
+      <TextField
+        id="feed-flow-input"
+        value={props.feedFlow.toString()}
+        label="Feed Flow"
+        unit={props.feedUnit}
+        type="number"
+        onChange={props.onChangeFeedFlow}
```
soofstad, line 23: Could also "unpack" the event here. So less has to be done in the parent.
`onChange={(event: ...) => onChangeFeedFlow(Number(event.target.value))}`

Alos, unpacking the "props" is common and nice to do.
`const {onChangeFeddFlow, onChangeFeedUnit} = props`

soofstad (approved):

web/src/components/feature/FeedFlowInput.tsx
mariuswinger: I like that even more!

ingeridhellen: I disagree regarding unpacking the props. personally, I like the fact that all props are grouped together, as it makes it easier to see where a value comes from. Also, you avoid the extra lines of unpacking.

soofstad: Not that important if the component is typed.
But if it's not, unpacking gives a nice overview of which props you have "to work with"

ingeridhellen: True, though in that case I would say the real issue was that the component was missing typing. Given that we always use typing, which we should, unpacking props is unnecessary and messy

ingeridhellen: I think the cleanest structure is something like this
```
type Props = {
  firstProp: string
  secondProp: string
}

const myComponent = (props: Props): void => {
  console.log(props.firstProp)
}

export default myComponent
```
I like the consistency of it. The input type is always `Props`, the inputs are always organized in the `props` object, there are only one component in the file and it is set as default, meaning that you can import like ìmport myComponent from './MyComponent'`

web/src/components/feature/FeedFlowInput.tsx
mariuswinger: Was not able to do this without defining a type. Do you approve of this solution?

web/src/components/feature/FeedFlowInput.tsx
soofstad: The exports are fine. Still, if there are only 2-4 props, I still like them defined in the interface :man_shrugging:

ingeridhellen: Why? Your suggestion allow for inconsistencies.

web/src/components/feature/FeedFlowInput.tsx
ingeridhellen: That might actually be even better

ingeridhellen (approved):

PR #70, opened 2022-10-11
feat: component for selecting components

mariuswinger:

Need to select components

Made component selector

### DIFF
--- a/web/src/components/feature/ComponentTable.tsx
+++ b/web/src/components/feature/ComponentTable.tsx
@@ -1,0 +1,46 @@
+import { EdsProvider, Table, TextField } from '@equinor/eds-core-react'
+
+export type TComponentEntry = {
+  componentId: string
+  altName: string
+  chemicalFormula: string
+}
+
+export const ComponentTable = ({
+  input,
+}: {
+  input: Array<TComponentEntry>
+}): JSX.Element => {
+  function createTableRows() {
+    return input.map((entry, index) => (
+      <Table.Row key={index}>
+        <Table.Cell
+          data-testid={`Component-${index}`}
+        >{`${entry.altName} (${entry.chemicalFormula})`}</Table.Cell>
+        <Table.Cell data-testid={`Feed Value (mol)-${index}`}>
+          <TextField
+            id={`Feed Value (mol)-${index}-input`}
+            placeholder={'0'}
+            type="number"
+          />
+        </Table.Cell>
+      </Table.Row>
+    ))
+  }
+
+  return (
+    <EdsProvider density={'compact'}>
+      <Table>
+        <Table.Head>
+          <Table.Row>
+            <Table.Cell key={`Component`}>{`Component`}</Table.Cell>
+            <Table.Cell
+              key={`Feed value (mol)`}
+            >{`Feed value (mol)`}</Table.Cell>
+          </Table.Row>
+        </Table.Head>
+        <Table.Body>{createTableRows()}</Table.Body>
+      </Table>
+    </EdsProvider>
+  )
+}
--- a/web/src/components/feature/ComponentSelector.tsx
+++ b/web/src/components/feature/ComponentSelector.tsx
@@ -1,0 +1,56 @@
+import { ComponentName, ComponentResponse } from '../../api/generated'
+import { Autocomplete } from '@equinor/eds-core-react'
+import { useState } from 'react'
+import styled from 'styled-components'
+import { ComponentTable, TComponentEntry } from './ComponentTable'
+
+const ComponentSelectorContainer = styled.div`
+  display: flex;
+  flex-direction: column;
+  row-gap: 4px;
+  max-width: 500px;
+`
+
+const ComponentTableContainer = styled.div`
+  display: flex;
+  flex-direction: column;
+  height: 400px;
+  overflow: auto;
+`
+
+function createOptions(components: ComponentResponse): Array<TComponentEntry> {
+  return Object.entries(components.components).map(
+    ([key, name]: [string, ComponentName]) => ({
+      componentId: key,
+      altName: name.altName,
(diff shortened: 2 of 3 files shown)


web/src/components/feature/ComponentTable.tsx
```diff
@@ -0,0 +1,83 @@
+import { ComponentResponse } from '../../api/generated'
+import { Autocomplete } from '@equinor/eds-core-react'
+import { useState } from 'react'
+import {
+  DynamicEditableTable,
+  TDynamicEditableTableInput,
+} from '../common/DynamicEditableTable'
+import styled from 'styled-components'
+
+type componentEntry = {
```
soofstad, line 10: Can we reuse/share the one from the other DynbamicTable?

web/src/components/common/DynamicEditableTable.tsx
```diff
@@ -0,0 +1,54 @@
...
+    })
+    return rows
+  }
+
+  return (
+    <EdsProvider density={'compact'}>
+      <Table>
+        <Table.Head>
+          <Table.Row>
+            {Object.keys(input).map((header) => (
```
soofstad, line 45: Hard code headers

web/src/components/common/DynamicEditableTable.tsx
```diff
@@ -0,0 +1,54 @@
+import { EdsProvider, Table, TextField } from '@equinor/eds-core-react'
+
+export type TDynamicEditableTableInput = {
+  // Keys will be used as headers. The array for each key will be the data in the column
+  [key: string]: { editable: boolean; values: Array<string> }
+}
+export const DynamicEditableTable = ({
```
soofstad, line 7: Dumb this down

soofstad (changes requested):

web/src/components/feature/ComponentTable.tsx
```diff
@@ -0,0 +1,83 @@
...
+          altName: x[1].altName,
+          chemicalFormula: x[1].chemicalFormula,
+        } as componentEntry)
+    )
+  }
+
+  return (
+    <ComponentSelectorContainer>
+      <ComponentTableContainer>
+        <DynamicEditableTable input={transformRawResponse(selectedEntries)} />
```
soofstad, line 68: Instead of calculating the input on every rerender, I think it would be cleaner to do the transformation once when a component is added to "selectedItems"

web/src/components/feature/ComponentTable.tsx
```diff
@@ -0,0 +1,83 @@
...
+}
+
+export const ComponentSelector = ({
+  components,
+}: {
+  components: ComponentResponse
+}) => {
+  const [selectedEntries, setSelectedEntries] = useState(Array<componentEntry>)
+
+  function createOptions(components: ComponentResponse) {
```
soofstad, line 52: Can and should be moved out of component if it's not refering to any "class" vars.

web/src/components/feature/ComponentTable.tsx
```diff
@@ -0,0 +1,83 @@
...
+
+    return options.map(
+      (x) =>
+        ({
+          componentId: x[0],
+          altName: x[1].altName,
+          chemicalFormula: x[1].chemicalFormula,
+        } as componentEntry)
+    )
+  }
```
soofstad, line 63: IMO it would be more idiomatic and cleaner to write it like this;
```
  function createOptions(components: ComponentResponse): any[] {
    return Object.entries(components.components).map(([key, value]: [string, ComponentName]) => ({
        componentId: key, altName: value.altName, chemicalFormula: value.chemicalFormula,
      }))

  }
```

soofstad (changes requested):

web/src/components/feature/ComponentTable.tsx
```diff
@@ -0,0 +1,50 @@
...
+
+  return (
+    <EdsProvider density={'compact'}>
+      <Table>
+        <Table.Head>
+          <Table.Row>
+            <Table.Cell key={`Component`}>{`Component`}</Table.Cell>
+            <Table.Cell
+              key={`Feed value (mol)`}
```
soofstad, line 42: No need to put "key" on components that are not part of a list

web/src/components/feature/ComponentTable.tsx
```diff
@@ -0,0 +1,50 @@
...
+          key={`Feed Value (mol)-${index}`}
+        >
+          <TextField
+            id={`Feed Value (mol)-${index}-input`}
+            placeholder={'0'}
+            type="number"
+          />
+        </Table.Cell>
+      </Table.Row>
+    ))
```
soofstad, line 32: Kind of strange having a function that only has another function and calls that.
I would rather write ComponentTable like this.
```tsx
const ComponentTable = (...) {
return <>
{input.map(() =>....)}
</>
}
```

web/src/components/feature/ComponentTable.tsx
```diff
@@ -0,0 +1,50 @@
...
+  input,
+}: {
+  input: Array<TComponentEntry>
+}): JSX.Element => {
+  function createTableRows() {
+    return input.map((entry, index) => (
+      <Table.Row key={index}>
+        <Table.Cell
+          data-testid={`Component-${index}`}
+          key={`Component-${index}`}
```
soofstad, line 19: No need to put "key" on components that are not part of a list

soofstad (changes requested):

web/src/components/feature/ComponentTable.tsx
mariuswinger: Agree. Slip-up.

mariuswinger: Not possible.

soofstad (approved):

PR #83, opened 2022-10-11
Style/main page

### LINKED ISSUE #77: Make MainPage layout as in sketches

ingeridhellen:

This pull request is needed because we want our main page to look good

Aligns main page to sketches. Since some features are missing, not everything is perfect yet, but this gets us at least halfway there

Closes #77

### DIFF
--- a/web/src/components/feature/PhaseTable.tsx
+++ b/web/src/components/feature/PhaseTable.tsx
@@ -31,6 +31,7 @@ export const PhaseTable = (props: {
         'Mercury Flow (g/d)',
       ]}
       rows={getRows(multiFlashResponse, cubicFeedFlow)}
+      density={'comfortable'}
     />
   )
 }
--- a/web/src/pages/Main.tsx
+++ b/web/src/pages/Main.tsx
@@ -7,6 +7,14 @@ import { AxiosResponse } from 'axios'
 import { ComponentResponse, MultiflashResponse } from '../api/generated'
 import { ComponentSelector } from '../components/feature/ComponentSelector'
 import { PhaseTable } from '../components/feature/PhaseTable'
+import styled from 'styled-components'
+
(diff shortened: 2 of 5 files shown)


web/src/components/feature/PhaseTable.tsx
```diff
@@ -31,6 +31,7 @@ export const PhaseTable = (props: {
         'Mercury Flow (g/d)',
       ]}
       rows={getRows(multiFlashResponse, cubicFeedFlow)}
+      density={'comfortable'}
```
mariuswinger, line 34: Do we want the tables to have different density?

web/src/pages/Main.tsx
```diff
@@ -8,6 +10,26 @@ import { ComponentResponse, MultiflashResponse } from '../api/generated'
 import { ComponentSelector } from '../components/feature/ComponentSelector'
 import { PhaseTable } from '../components/feature/PhaseTable'
 
+const Results = styled.div`
+  display: flex;
+  justify-content: flex-start;
+  gap: 50px;
+  align-items: center;
+  flex-wrap: wrap;
```
mariuswinger, line 18: This behaves a bit weird when resizing the window (i.e. dev tools). It wraps when the window becomes to small to have both tables on the same row as expected, but when expanding it puts them on the same row until some point where it wraps, even though there is enough space, before putting them on the same row again.
I think it has something to do with min-width: 1200px in Container.

mariuswinger (approved):

mariuswinger:
Would be nice to see this with a rebase so the tables are generated when clicking run.

web/src/components/feature/PhaseTable.tsx
ingeridhellen: According to the sketches they should

Issue #58, opened 2022-09-29
CI: Automatic dependency bumping (snyk, dependabot?)


soofstad:

soofstad:
dependabot?
Remove safety as dev-dependency

PR #87, opened 2022-10-12
ci: Remove safety from dev-deps

mariuswinger:

Remove safety from dev-deps in pyproject.toml

#58


ingeridhellen (changes requested):
Should be removed from poetry.lock as well

ingeridhellen (approved):

PR #103, opened 2022-10-13
Feature/place componentselector

### LINKED ISSUE #84: Move ComponentSelector into Dialog

ingeridhellen:

This pull request is needed because the ComponentSelector should be inside the Dialog

* Moves ComponentSelector into the dialog
* Adjust dialog width and styling to fit the new content
* Change order of autocomplete and table inside ComponentSelector

Closes #84

### DIFF
--- a/web/src/components/feature/FluidDialog.tsx
+++ b/web/src/components/feature/FluidDialog.tsx
@@ -4,7 +4,10 @@ import { ComponentSelector } from './ComponentSelector'
 import { ComponentResponse } from '../../api/generated'
 
 const WideDialog = styled(Dialog)`
-  min-width: 700px;
+  width: auto;
+  @media (min-width: 900px) {
+    min-width: 800px;
+  }
 `
 
 const FluidPackageForm = styled.div`
--- a/web/src/components/feature/FluidDialog.tsx
+++ b/web/src/components/feature/FluidDialog.tsx
@@ -1,11 +1,15 @@
 import { Button, Dialog, TextField } from '@equinor/eds-core-react'
+import { ComponentSelector } from './ComponentSelector'
+import { ComponentResponse } from '../../api/generated'
 
 export const FluidDialog = ({
   open,
   onClose,
+  components,
 }: {
   open: boolean
   onClose: () => void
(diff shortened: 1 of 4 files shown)


web/src/components/feature/FluidDialog.tsx
```diff
@@ -1,28 +1,55 @@
...
+  display: flex;
+  flex-flow: row wrap;
+  gap: 30px;
+`
+
+const FirstColumn = styled.div`
+  display: flex;
+  flex-flow: column nowrap;
+  gap: 30px;
+`
```
mariuswinger, line 20: The styling is not quite right, I think. When selecting "2-2-dimethyl-butane(neohexane) (22-dm-C4)" (the longest component => ComponentSelector becomes wider), the selector jumps beneath the text fields.

web/src/components/feature/FluidDialog.tsx
```diff
@@ -33,6 +60,6 @@ export const FluidDialog = ({
         </Button>
         <Button type="submit">Save</Button>
       </Dialog.Actions>
-    </Dialog>
+    </WideDialog>
```
mariuswinger, line 63: I think (some of) the buttons should be on the right side?

ingeridhellen: Yes, but that is outside the scope of this task

soofstad (approved):
Addition to the other remarks.
Can we make it so the "Select components" component does not have focus on dialog opening?
Point is to not have the drop down open on Dialog opening

Same goes for the drop down for selecting "Fluid package" btw.

ingeridhellen:
> Fluid package



> Addition to the other remarks. Can we make it so the "Select components" component does not have focus on dialog opening? Point is to not have the drop down open on Dialog opening
> 
> Same goes for the drop down for selecting "Fluid package" btw.

Nope, this is a bug in the library used by EDS. However, it should apparently only happen when running in strictmode, which is why it doesn't occur in dev. See https://github.com/equinor/design-system/issues/2369 for details

mariuswinger (approved):

PR #105, opened 2022-10-13
fix: make it possible to input negative numbers

### LINKED ISSUE #104: web: allow negative values to be typed into text field

mariuswinger:

User might have negative values to enter. This was not possible/very hard previously.

Remove value prop and substitute with placeholder.

Closes #104

### DIFF
--- a/web/src/components/feature/CalculateFluid.tsx
+++ b/web/src/components/feature/CalculateFluid.tsx
@@ -33,10 +33,9 @@ export const CalculateFluid = ({
   setResult: (result: MultiflashResponse) => void
 }) => {
   const [isOpen, setIsOpen] = useState<boolean>(false)
-  const [temperature, setTemperature] = useState(15)
-  const [pressure, setPressure] = useState(1)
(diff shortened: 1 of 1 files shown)


soofstad (approved):

web/src/components/feature/CalculateFluid.tsx
```diff
@@ -101,7 +100,7 @@ export const CalculateFluid = ({
             <TextField
               id="temperature-input"
               min={-273.15}
-              value={temperature.toString()}
+              placeholder={temperature.toString()}
```
ingeridhellen, line 103: What about instead doing
```suggestion
              defaultValue={temperature.toString()}
```
?

ingeridhellen (approved):

web/src/components/feature/CalculateFluid.tsx
mariuswinger: Yes, that is better. Was looking for it, but couldn't find it.

PR #109, opened 2022-10-14
ci: disable eslint for building

soofstad:
Disables eslint for running and building the web app. Also fixes eslint pre-commit hook to not show warning on ignored files, and to actually fix the code when run.

The reasoning is that a linting tool should not stop the app from building. The linting will still be checked in the tests.

Note: Open for debate on this one...

### DIFF
--- a/web/package.json
+++ b/web/package.json
@@ -38,8 +38,8 @@
     "typescript": "^4.7.4"
   },
   "scripts": {
-    "start": "react-scripts start",
-    "build": "react-scripts build",
+    "start": "ESLINT_NO_DEV_ERRORS='true' react-scripts start",
+    "build": "DISABLE_ESLINT_PLUGIN=true react-scripts build",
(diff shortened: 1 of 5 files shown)


web/package.json
```diff
@@ -38,8 +38,8 @@
     "typescript": "^4.7.4"
   },
   "scripts": {
-    "start": "react-scripts start",
-    "build": "react-scripts build",
+    "start": "DISABLE_ESLINT_PLUGIN=true react-scripts start",
```
ingeridhellen, line 41: Do we really want it disabled on yarn start?

ingeridhellen (approved):

web/package.json
soofstad: This is a per-developer preference. But you'r right, most will likely like to keep this. As long as it still compiles with eslint errors. I will reenable that.

PR #111, opened 2022-10-14
feat: multiflash input update from FluidDialog

### LINKED ISSUE #110: web: update multiflash input components from FluidDialog

mariuswinger:
Multiflash computation should take values from ComponentTable

Adds states for collecting inputs from table and handle to update when save button is clicked. According to the figma sketches, the save button will be renamed to "ok" and/or replaced with "update package". 

Closes #110

### DIFF
--- a/web/src/components/feature/FluidDialog.tsx
+++ b/web/src/components/feature/FluidDialog.tsx
@@ -1,8 +1,8 @@
 import styled from 'styled-components'
 import { Button, Dialog, TextField } from '@equinor/eds-core-react'
 import { ComponentSelector } from './ComponentSelector'
-import { ComponentName, ComponentResponse } from '../../api/generated'
-import { TFeedComponentRatios } from '../../pages/Main'
+import { ComponentResponse } from '../../api/generated'
+import { TComponentComposition } from '../../pages/Main'
 import { useState } from 'react'
 import { demoComponentInput, demoFeedComponentRatios } from '../../constants'
 
@@ -25,6 +25,13 @@ const FirstColumn = styled.div`
   gap: 30px;
 `
 
+export type TComponent = {
+  componentId: string
+  altName: string
+  chemicalFormula: string
+  value: number
+}
+
 export type TComponentInput = {
   [componentId: string]: {
     altName: string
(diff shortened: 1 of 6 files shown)


ingeridhellen (changes requested):
We're doing a lot of mapping of components, which makes the code really confusing. In addition, many of the component objects/arrays have similar names. Could you try refactoring this? I think perhaps we should stop using objects and instead use arrays like
```
const components = [{componentId. name, formula, ratio}]
```

web/src/components/feature/FluidDialog.tsx
```diff
@@ -22,15 +25,84 @@ const FirstColumn = styled.div`
...
+}
+
+function normalizeFeedComponentRatios(
+  feedComponentRatios: TFeedComponentRatios
+): TFeedComponentRatios {
+  // Normalize feedComponentRatios if necessary, else return
+  const normalizedValues: Array<number> = normalizeArray(
+    Object.values(feedComponentRatios)
+  )
+  if (normalizedValues !== Object.values(feedComponentRatios)) {
```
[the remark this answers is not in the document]
mariuswinger: pythonism. Will fix.

web/src/components/feature/FluidDialog.tsx
```diff
@@ -61,7 +136,17 @@ export const FluidDialog = ({
         <Button variant="outlined" onClick={onClose}>
           Cancel
         </Button>
-        <Button type="submit">Save</Button>
+        <Button onClick={handleSubmit}>Save</Button>
+        <Button
+          // TODO: Demo button to remove when done testing
```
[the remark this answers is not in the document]
mariuswinger: Useful if you want some synthetic data without typing it in, but otherwise completely unnecessary.

web/src/components/feature/FluidDialog.tsx
```diff
@@ -22,15 +25,93 @@ const FirstColumn = styled.div`
...
+  componentComposition: TComponentComposition
+): TComponentComposition {
+  // Normalize feedComponentRatios if necessary, else return
+  const normalizedValues: Array<number> = normalizeArray(
+    Object.values(componentComposition)
+  )
+  if (
+    JSON.stringify(normalizedValues) !==
+    JSON.stringify(Object.values(componentComposition))
+  ) {
```
ingeridhellen, line 76: A bit overcomplicated. Perhaps just
```suggestion
  const values = Object.values(componentComposition)
  // Normalize feedComponentRatios if necessary, else return
  const normalizedValues: Array<number> = normalizeArray(values)
  if (normalizedValues !== values) {
```

ingeridhellen (approved):

web/src/components/feature/FluidDialog.tsx
mariuswinger: Not sure the comparison would work? I think(?) javascript compares arrays as objects meaning that it will return false because the two arrays have different memory addresses.

ingeridhellen: I sent you an image of my console, and then it worked perfectly. But this PR blocks a lot, so I'll just merge it anyway

PR #113, opened 2022-10-18
fix: add precision formatting for tables

### LINKED ISSUE #98: web(fix): Presicion on PhaseTable should be scientific if < E-04

### LINKED ISSUE #99: web(fix): Precision on MoleTable should be 2 decimals

mariuswinger:
Formatting in the tables are currently haywire.

Table entries are now formatted.

Closes #98, closes #99

### DIFF
--- a/web/src/tableUtils.ts
+++ b/web/src/tableUtils.ts
@@ -1,0 +1,20 @@
+/**
+ * format a number to scientific notation.
+ *
+ * @param {number} number - Number to format
+ * @param {number} precision - number of significant digits in scientific notation
+ * @param {number} threshold - limit on when to convert a number to scientific notation
+ * @returns {string} number formatted in scientific notation
+ */
+export function formatNumber(
+  number: number,
+  precision = 4,
+  threshold = 4
+): string {
+  if (number < 1 / 10 ** (threshold - 1) || number > 10 ** threshold) {
+    return number.toExponential(precision).replace('e', 'E')
+  } else {
+    // Set precision to given number of digits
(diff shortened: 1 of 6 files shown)


web/src/tableUtils.ts
```diff
@@ -0,0 +1,12 @@
+export function formatNumber(
```
ingeridhellen, line 1: Could you add a docstring on this one? On the format
```suggestion
/**
 *
 * @param {string} param0
 * @returns
 */
export function formatNumber(
```
It is not obvious what the threshold or precision is, and what's their default values

web/src/tableUtils.ts
```diff
@@ -0,0 +1,12 @@
...
+  number: number,
+  precision?: number,
+  threshold?: number
+): string {
+  threshold = threshold ? threshold : 4
+  if (number < 10 ** -threshold || number > 10 ** threshold) {
+    return number.toExponential(precision ? precision : 4).replace('e', 'E')
+  } else {
+    return number.toPrecision(precision ? precision : 4).toString()
```
ingeridhellen, line 10: In JS, you can take default values like this
`function myFunc(myBool: boolean | undefined = true)`
Also, instead of `precision ? precision : 4` you can do `precision ?? 4`, though if you use default values this it not necessary anymore

ingeridhellen (approved):

web/src/tableUtils.ts
mariuswinger: Good suggestion. Will do.

PR #114, opened 2022-10-18
refactor: Refactor ratio normalization

ingeridhellen:

This pull request is needed to make the code easier to read and more consistent

Rewrites how we normalize component composition

### DIFF
--- a/web/src/components/feature/FluidDialog.tsx
+++ b/web/src/components/feature/FluidDialog.tsx
@@ -52,35 +52,24 @@ function convertComponentResponseToComponentInput(
   )
 }
 
-function normalizeArray(array: Array<number>): Array<number> {
-  // Normalize array if not "close enough" (|sum - 1| > 0.01) to 1
+function getNormalizationFactor(array: Array<number>): number {
   const arraySum: number = array.reduce(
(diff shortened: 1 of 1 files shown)


web/src/components/feature/FluidDialog.tsx
```diff
@@ -52,35 +52,24 @@ function convertComponentResponseToComponentInput(
...
 }
 
 function normalizeComponentComposition(
   componentComposition: TComponentComposition
 ): TComponentComposition {
-  // Normalize feedComponentRatios if necessary, else return
-  const normalizedValues: Array<number> = normalizeArray(
-    Object.values(componentComposition)
+  const factor = getNormalizationFactor(Object.values(componentComposition))
+  return Object.fromEntries(
```
ingeridhellen, line 67: @mariuswinger I chose to create a new object instead of rewriting the existing one. This is best practice in React, since rewriting some part of an object or array will not trigger a rerender, while creating a new one that replaces the old, will. Not strictly necessary in this case, but good to be in the habbit

@StigNil An example of where using arrays would have been much smoother

mariuswinger (approved):

rikkebl (approved):

PR #121, opened 2022-10-21
feature: load saved package when selected

soofstad:

Should be able to reuse saved packages

- Rename "CalculateFluid" component to "Input"
- Refactor stored packages to be a dict instead of a list with packages
- Load the package selected in the dropdown

none, but needed for #120

### DIFF
--- a/web/src/components/common/Header.tsx
+++ b/web/src/components/common/Header.tsx
@@ -72,6 +72,13 @@ export const Header = (): JSX.Element => {
           >
             Logout
           </Button>
+          <Button
+            onClick={() => localStorage.removeItem('packages')}
+            color={'secondary'}
+            variant={'outlined'}
+          >
+            Reset application
+          </Button>
         </Popover.Actions>
       </Popover>
       <Popover open={isAboutOpen} anchorEl={infoRefElement.current}>
--- a/web/src/components/feature/CalculationInput.tsx
+++ b/web/src/components/feature/CalculationInput.tsx
@@ -1,5 +1,6 @@
 import {
   Autocomplete,
+  AutocompleteChanges,
   Button,
   Progress,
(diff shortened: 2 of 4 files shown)


web/src/components/feature/Input.tsx
```diff
@@ -144,6 +160,7 @@ export const CalculateFluid = ({
         packages={packages}
         setPackages={setPackages}
       />
+      {/*<pre>{JSON.stringify(componentComposition, null, 2)}</pre>*/}
```
ingeridhellen, line 163: For testing?

web/src/components/feature/Input.tsx
```diff
@@ -33,7 +34,7 @@ const FluidPackage = styled(FlexContainer)`
   }
 `
 
-export const CalculateFluid = ({
+export const Input = ({
```
ingeridhellen, line 37: I'm not a fan of naming things input, because it's to generic. By looking at only the name, I would not be able to guess if this was a dropdown, a textfield, our package input dialog or our main input area. And text searches would get a number of falls positives. Could we come up with something more specific? Like CalculationInput perhaps? InputCard, maybe, though I'm not sure the name should refer to a specific component we've used. CalculationConfigs? InputAndRun?

ingeridhellen (changes requested):

web/src/components/feature/Input.tsx
soofstad: CalculationInput is good :slightly_smiling_face:

web/src/components/feature/Input.tsx
soofstad: yes, removed

ingeridhellen (approved):

Issue #118, opened 2022-10-19
Rename onClose for Dialog


ingeridhellen:

soofstad:
closed by #122

PR #122, opened 2022-10-24
fix: only preSelected components was selectable

soofstad:

Fixes a bug where only the 12 preSelected somponents where selectable in the ComponentSelector.

Also some refactor/minor changes:

 - react hot reload now works on port 80 (WDS_SOCKET_PORT env var)
 - disable run button when no components are selected
 - disable save button when no packageName is set
 - rename FluidDialog 'open' and 'onClose' to 'isOpen' and 'close'
 - render some helpful text where the results will come

none, and resoles #118

### DIFF
--- a/web/package.json
+++ b/web/package.json
@@ -38,7 +38,7 @@
     "typescript": "^4.7.4"
   },
   "scripts": {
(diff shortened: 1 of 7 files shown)


ingeridhellen (approved):
Looks good

mariuswinger (approved):

Issue #74, opened 2022-10-11
saves a new fluid/overwrites the selected fluid (confirm choice?)


soofstad:

soofstad:
See https://github.com/soofstad/react-oauth2-pkce/blob/main/src/Hooks.tsx for Localstorage hook

mariuswinger:
Looks to be done?

PR #124, opened 2022-10-24
Refactor/types

ingeridhellen:

This pull request is needed to make the code easier to read

- Create new types TComponentNames and TComponentRatios that will be used instead of ComponentResponse and TComponentComposition.
- Rename variables
- Simplify CalculationInput by using package as list, since that is more compatible with Autocomplete. Note that it might be preferable to map it to an object before passing into FluidDialog, but at the moment there is no need.

Refs #74


web/src/pages/Main.tsx
```diff
@@ -34,30 +34,43 @@ const DividerWithLargeSpacings = styled(Divider)`
...
     unit: 'Sm3/d',
     value: 1000,
   })
-  const [componentComposition, setComponentComposition] =
-    useState<TComponentComposition>()
-  const [result, setResult] = useState<MultiflashResponse>()
+  const [usedComponentRatios, setUsedComponentRatios] = useState<
+    TComponentRatios | undefined
+  >()
+  const [result, setResult] = useState<MultiflashResponse>({
```
soofstad, line 47: IMO its better to leave this to undefined instead of setting a mock result.
Typescript will then help us to not do anything with the result before it has some actual value

web/src/setupTests.ts
```diff
@@ -60,39 +61,37 @@ export const mockMultiflashResponse: MultiflashResponse = {
...
-    '608': { chemicalFormula: 'benzene', altName: 'Benzene' },
-    '701': { chemicalFormula: 'nC7', altName: 'n-heptane' },
-    '707': { chemicalFormula: 'cy-C7', altName: 'Cy-heptane' },
-    '710': { chemicalFormula: 'toluene', altName: 'Toluene' },
-    '801': { chemicalFormula: 'nC8', altName: 'n-octane' },
-    '806': { chemicalFormula: 'cy-C8', altName: 'Cy-octane' },
-    '809': { chemicalFormula: 'm-xylene', altName: 'M-xylene' },
-    '901': { chemicalFormula: 'nC9', altName: 'n-nonane' },
-    '1016': { chemicalFormula: 'nC10', altName: 'n-decane' },
+export const mockComponentProperties: TComponentProperties = {
```
soofstad, line 64: Why move away from using the generated types?
There are some changes incoming, and using those types would help us detect errors

soofstad (changes requested):
Good stuff! :+1: 
A few coments

web/src/setupTests.ts
soofstad: We can also extend these generated types if needed

web/src/pages/Main.tsx
ingeridhellen: Right, yes, I agree completely. It seems to have reappeared when solving a merge conflict.

web/src/setupTests.ts
ingeridhellen: True, but the generated type has a name (ComponentResponse) that carries little meaning IMO, and is also in conflict with how we normally name types. Also, I think backend should be able to change its types without us having to fix the entire frontend. With this code Marius could rename altName to something else, and frontend would only need to be changed on one line. Finally, by having our own type, we don't have to do `ComponentResponse.component` everywhere, so the code becomes more compact

web/src/components/feature/FluidDialog.tsx
```diff
@@ -45,68 +32,54 @@ function getNormalizationFactor(array: Array<number>): number {
...
+      setPackageDescription(selectedPackage.description)
+      setComponentRatios(selectedPackage.components)
+    }
+  }, [selectedPackage])
 
   const savePackage = () => {
-    const componentComposition = getComponentComposition()
-    setComponentComposition(normalizeComponentComposition(componentComposition))
-    setPackages({
+    setPackages([
```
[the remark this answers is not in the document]
ingeridhellen: That's a good point. But Autocomplete needs a list, so I propose using array in calculationinput and object in fluiddialog

soofstad (approved):
Nice!

PR #133, opened 2022-10-25
feat: add conversion of feed flow between units

### LINKED ISSUE #80: Add support for feed in kg/d
To feedFlow input

mariuswinger:
Add support for conversion of feed units.

User can now convert between units and the conversion will happen in the UI. Unit switch is disabled until a composition feed is selected.

Closes #80

### DIFF
--- a/web/src/pages/Main.tsx
+++ b/web/src/pages/Main.tsx
@@ -9,7 +9,7 @@ import { AxiosResponse } from 'axios'
 import { ComponentResponse, MultiflashResponse } from '../api/generated'
 import { PhaseTable } from '../components/feature/results/PhaseTable'
 import { MercuryWarning } from '../components/feature/results/MercuryWarning'
-import { TComponentProperties, TComponentRatios, TFeedFlow } from '../types'
+import { TComponentProperties, TComponentRatios } from '../types'
 
 const Results = styled.div`
   display: flex;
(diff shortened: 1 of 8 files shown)


web/src/types.ts
```diff
@@ -12,4 +12,5 @@ export type TPackage = {
   name: string
   description: string
   components: TComponentRatios
+  molecularWeight: number
```
ingeridhellen, line 15: Isn't molecular weight a constant of the chemical compound? If so, I don't think it makes sense to store it as part of the package

web/src/pages/Main.tsx
```diff
@@ -37,10 +37,8 @@ export const MainPage = (props: { mercuryApi: MercuryAPI }): JSX.Element => {
   const [componentProperties, setComponentProperties] =
     useState<TComponentProperties>()
   const [isLoading, setIsLoading] = useState<boolean>(true)
-  const [feedFlow, setFeedFlow] = useState<TFeedFlow>({
```
ingeridhellen, line 40: Does that mean we can remove the TFeedFlow type?

web/src/types.ts
ingeridhellen: Ah, no, okay, I get it. Could we perhaps call it something else then? combinedMolecularWeight, molecularWeightSum or something like that? I see you call it feedMolecularWeight in FeedFlowInput, but I'm not sure that is better. Best if it's called the same everywhere, though

web/src/components/feature/FeedFlowInput.tsx
```diff
@@ -1,40 +1,72 @@
...
+import { molePerStandardCubicMeter } from '../../constants'
+import { useState } from 'react'
 
 const FlexContainer = styled.div`
   display: flex;
   gap: 16px;
   max-width: 250px;
 `
 
+function convertCubicFeedFlowValue(
```
ingeridhellen, line 12: I know I send in cubicFeedFlow. The question is what I get out. Perhaps rename to `getMassFeedFlow` or something like that, and similarly for the function below

web/src/types.ts
mariuswinger: I agree. Will rename it.

web/src/pages/Main.tsx
mariuswinger: Yes. We assume the feed flow is cubic everywhere and just convert/display it as mass flow if the switch is clicked.

web/src/components/feature/FeedFlowInput.tsx
mariuswinger: Yeah. Was trying to avoid a very long name, but probably better clarity that way.

PR #134, opened 2022-10-25
fix: normalized feed was not correct

mariuswinger:
There was a bug in the computation of the normalized feed ratios in the response model.

The bug is dead. Also removed normalization from web and updated the mole table.


api/src/entities/Multiflash.py
```diff
@@ -57,12 +57,14 @@ def feed_composition(self) -> List[float]:
     def number_of_components(self) -> int:
         return len(self.component_ids)
 
+    # @property
+    # def normalized_feed_composition(self) -> Dict[str, float]:
+    #     normalized_array = np.array(self.feed_composition) / np.sum(self.feed_composition)
+    #     return {component_id: feed_ratio for component_id, feed_ratio in zip(self.component_ids, normalized_array)}
```
soofstad, line 63:
```suggestion
```

PR #136, opened 2022-10-25
feat: Add edit package button

### LINKED ISSUE #24: Add button to edit existing fluid package

ingeridhellen:

This pull request is needed to allow users to edit existing packages in an intuitive way

Adds an edit button and logic to go with it.

Closes #24

### DIFF
--- a/web/src/components/feature/calculation_input/CalculationInput.tsx
+++ b/web/src/components/feature/calculation_input/CalculationInput.tsx
@@ -58,6 +58,7 @@ export const CalculationInput = ({
   const [calculating, setCalculating] = useState<boolean>(false)
   const [packages, setPackages] = useLocalStorage<TPackage[]>('packages', [])
   const [selectedPackage, setSelectedPackage] = useState<TPackage | undefined>()
+  const [editablePackage, setEditablePackage] = useState<TPackage | undefined>()
 
   const calculate = (
     <Button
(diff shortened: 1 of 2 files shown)


web/src/components/feature/calculation_input/CalculationInput.tsx
```diff
@@ -95,6 +96,18 @@ export const CalculationInput = ({
     </Button>
   )
 
+  const savePackage = (newPackage: TPackage) => {
+    let newPackages = [...packages]
```
mariuswinger, line 100: Can this variable be renamed to allPackages or something along those lines? Very hard to see the difference between newPackage and newPackage**s**.

web/src/components/feature/calculation_input/CalculationInput.tsx
```diff
@@ -95,6 +96,18 @@ export const CalculationInput = ({
...
+    if (editablePackage !== undefined) {
+      newPackages = newPackages.filter((x) => x.name !== editablePackage.name)
+    }
```
mariuswinger, line 103: Is this included so that packages can be renamed? Or are there any other reasons to why we need a separate state for editablePackage? Naively I would have thought it is as easy as saying that a package can be edited if it is non-empty.

ingeridhellen: editablePackage does two things. It tells the dialog what mode it is in (edit or new) based on whether editablePackage is undefined or not. If it is in edit mode, ie editablePackage is not undefined, then editablePackage contains the package we want to edit.
I tried finding a better name for this parameter, but could not. If you have any, please tell

web/src/components/feature/calculation_input/CalculationInput.tsx
ingeridhellen: True. And the name isn't even correct. Fixed now

mariuswinger (approved):

PR #151, opened 2022-10-26
feat: add units to PhaseTable

rikkebl:
Get units on phase table.

Add units on phase table (mass concentration). 

#138

### DIFF
--- a/web/src/components/feature/results/PhaseTable.tsx
+++ b/web/src/components/feature/results/PhaseTable.tsx
@@ -1,6 +1,6 @@
 import { DynamicTable } from '../../common/DynamicTable'
 import { MultiflashResponse } from '../../../api/generated'
-import { formatNumber } from '../../../tableUtils'
+import { formatNumber, getCorrectUnit } from '../../../tableUtils'
 import {
   mercuryMolecularWeight,
   molePerStandardCubicMeter,
@@ -17,9 +17,10 @@ function getRows(
     ([phase, values], index) => [
       phase,
       formatNumber(values['percentage']),
-      formatNumber(values['mercury']),
-      formatNumber(mercury[index]),
-      formatNumber(phPhaseFlowFactor * values['percentage'] * mercury[index]),
+      formatNumber(values['mercury']) + getCorrectUnit(phase),
+      formatNumber(mercury[index]) + ' mol/mol',
+      formatNumber(phPhaseFlowFactor * values['percentage'] * mercury[index]) +
+        ' g/d',
     ]
   )
 }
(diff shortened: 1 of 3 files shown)


mariuswinger:
imo looks a bit wonky with units in a single column. Would rather have units in all of them.

web/src/components/feature/results/PhaseTable.tsx
```diff
@@ -34,7 +34,7 @@ export const PhaseTable = (props: {
         'Phases',
         'Ratio',
         'Mass Concentration',
-        'Mole Concentration',
+        'Mole Concentration (mol)',
```
mariuswinger, line 37: Not sure if the labels were correct. I think they should reflect that the numbers are either the concentration of mercury measured volumetric or by mole. "Mercury concentration" fits for both, but is bad since it does not separate enough. "Mercury Mass/Mole concentration" is very describing, but perhaps too long. Not sure what to go for here ...

web/src/tableUtils.ts
```diff
@@ -18,3 +18,7 @@ export function formatNumber(
     return number.toPrecision(precision).toString()
   }
 }
+
+export function getCorrectUnit(phase: string): string {
+  return phase === 'Vapor' ? ' mg/Sm3' : ' mg/Kg'
```
mariuswinger, line 23: mg is milligrams in the SI system. It should be micrograms, i.e. μg (or "microgram").

rikkebl: ups, i'll fix

web/src/components/feature/results/PhaseTable.tsx
rikkebl: I think "Mercury Mass/Mole concentration" is too long. But not sure what the best header for this column is. Can remove mol, and try to put units in all colums, to see if that looks better.

ingeridhellen (approved):

web/src/components/feature/results/PhaseTable.tsx
soofstad: I suggest;
- Phases
- Fractions
- Concentration (mol)
- Concentration (ug)

rikkebl: Stig Oskar's suggestion:
image

or with units in all columns: 
image

mariuswinger: Oh, and I think the k in kg should be lower case.

rikkebl: image

mariuswinger: I'm fine with this. 💁‍♂️ Can always be changed later on.

rikkebl: Could leave it like this for now, and than change later if we find a better solution

PR #152, opened 2022-10-26
fix: Rerender dialog on opening

### LINKED ISSUE #145: web: non-default components are not shown in dialog upon editing
1. Press New
2. Add some components that are not added to table as default
3. Give these components a value
4. Press Save
5. Press Edit

The components are not shown. All components with a value should be shown as default

ingeridhellen:

This pull request is needed because we'd want the components with value to show in dialog when opening, not the defaults

- Dialog becomes more predictable if the entire dialog rerenders on opening. I've therefore wrapped it in a conditional and removed the isOpen input parameter
- I found a more intuitive solution the new/edit mode that allows me to remove editablePackage from CalculationInput. This also removes the danger of a race condition between isOpen and editablePackage
- Set autocomplete initials to depend on the componentRatios object

Closes #145


soofstad (approved):
Nice. Like these changes a lot :)

rikkebl (approved):

PR #153, opened 2022-10-26
style: Increase CalculationInput width

ingeridhellen:

This pull request is needed because of styles are good

Fixed card widht


rikkebl (approved):
Impressive work!

Issue #157, opened 2022-10-27
fix(web): feed value input fields don't accept writing 0.01


ingeridhellen:
How to reproduce
1. Press New
2. In one of the Feed value (mol) textfields try writing 0.01
When you try to write the second zero the "." disappears

Solution: Try replacing TextField with [Input](https://eds-storybook-react.azurewebsites.net/?path=/docs/inputs-input--introduction)

PR #158, opened 2022-10-27
fix: use Input in ComponentTable

rikkebl:
To accept decimalnumbers on format 0.01

Changed from Textfield to Input in ComponentTable. 
Removed toString(). 

#157


mariuswinger (approved):
Nice! Should we do this for the other number input fields as well?

rikkebl:
Tried to use it on temperature and pressure, but the Input don't support the warning helpertext.

soofstad:
Could add some custom "helperText" then. But not sure how bad these TextFields behave right now, if its worth it or not

rikkebl:
I'll just merge this now, and we could discuss that later

PR #165, opened 2022-11-01
fix: app handles zero mercury

### LINKED ISSUE #142: Zero mercury crashes page
Reproduction:
1. Create a new fluid package with demo data, except for mercury, which is set to 0
2. Run calculations with default temp and pressure

The problem seems to be that the componentFractions does not have the mercury key,. This causes PhaseTable to crash, since mole concentration and mercury flow cannot be calculated. Not sure how to best solve this though

mariuswinger:
App crash for zero mercury.

Defaults zero mercury to array of zeros. Could be discussed how meaningful the output is and if it is perhaps better to give a warning about how mercury is zero and disallow computation?

Closes #142


soofstad:
Dont handle, deny creating packages with zero mercury

ingeridhellen (approved):

PR #176, opened 2022-11-03
feat: add version number to header

### LINKED ISSUE #166: Add version.txt and commit sha to "About" in web page
Example: echo $(git log -n 1 --format=format:'%h% %>(12) %cs %>(12,mtrunc)%aN%  %s %D') > ./web/packages/home/public/version.txt

mariuswinger:
Displays the current version number in the header.

Add axios call to read version.txt in public folder. 

Closes #166

Could not suppress the 404 in console if no version.txt is found. Should not be a problem for non-local setups as the version.txt exists in the docker-image and in ci.

### DIFF
--- a/.github/workflows/publish.yaml
+++ b/.github/workflows/publish.yaml
@@ -78,7 +78,7 @@ jobs:
           echo "Tagging with ${{ inputs.image-tag }}"
           echo ${{ secrets.GITHUB_TOKEN }} | docker login $IMAGE_REGISTRY -u $GITHUB_ACTOR --password-stdin
           docker pull $API_IMAGE
-          echo "$(git log -n 1 --format=format:'%h% %>(12) %cs %aN%  %s %d')"
+          printf "$(git log -n 1 --format=format:'%h% %>(12) %cs %aN%  %s %d')" > ./api/src/version.txt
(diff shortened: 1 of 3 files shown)


.github/workflows/publish.yaml
```diff
@@ -79,7 +78,7 @@ jobs:
           echo "Tagging with ${{ inputs.image-tag }}"
           echo ${{ secrets.GITHUB_TOKEN }} | docker login $IMAGE_REGISTRY -u $GITHUB_ACTOR --password-stdin
           docker pull $API_IMAGE
-          echo $(git log -n 1 --format=format:'%h% %>(12) %cs %>(12,mtrunc)%aN%  %s %D') > ./api/src/version.txt
+          echo "$(git log -n 1 --format=format:'%h% %>(12) %cs %aN%  %s %d')"
```
soofstad: Not much value echoing this in the CI only. Should be put into a file that the API writes out to logs at startup

soofstad (changes requested):

.github/workflows/publish.yaml
mariuswinger: Then I misunderstood something. Fixed.

soofstad (approved):

PR #182, opened 2022-11-04
feature: warn user before overwriting existing package

### LINKED ISSUE #161: web(feat): Warn on overriding existing package
The user should get a confirmation dialog when trying to save a new package with a name collision with an existing package

soofstad:

User should be warned before any potential destructive actions

Prompt the user for a confirmation before overriding a FluidPackage with a new one

closes #161

### DIFF
--- a/web/src/components/feature/calculation_input/CalculationInput.tsx
+++ b/web/src/components/feature/calculation_input/CalculationInput.tsx
@@ -95,7 +95,7 @@ export const CalculationInput = ({
     const packageNameConflict = oldPackages.find(
       (x) => x.name === newPackage?.name
     )
-    if (mode === 'new' && packageNameConflict) {
+    if (packageNameConflict) {
       const overwrite: boolean = window.confirm(
         `A package named '${newPackage?.name}' already exists.\nOk to overwrite?`
       )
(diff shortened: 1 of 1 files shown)


web/src/components/feature/calculation_input/CalculationInput.tsx
```diff
@@ -91,6 +91,16 @@ export const CalculationInput = ({
     if (mode === 'edit' && selectedPackage !== undefined) {
       oldPackages = oldPackages.filter((x) => x.name !== selectedPackage.name)
     }
+
+    const packageNameConflict = oldPackages.find(
+      (x) => x.name === newPackage?.name
+    )
+    if (mode === 'new' && packageNameConflict) {
```
ingeridhellen, line 98: What if you're renaming an existing package to a name that already exists? Don't you want a warning then before overwriting the other package?

ingeridhellen (changes requested):

web/src/components/feature/calculation_input/CalculationInput.tsx
soofstad: Good point!

ingeridhellen (approved):

PR #189, opened 2022-11-07
feature(web): be able to use existing packages as templates

### LINKED ISSUE #188: Feat(New package): Select package as template in FluidDialog

soofstad:

Users want to create fluid packages that are almost similar, templates are a solution for that.

Adds a dropdown for selecting packages as a template when creating a new, or editing a package

Screenshot_20221107_101617

closes #188

### DIFF
--- a/web/src/components/feature/calculation_input/CalculationInput.tsx
+++ b/web/src/components/feature/calculation_input/CalculationInput.tsx
@@ -156,6 +156,7 @@ export const CalculationInput = ({
           close={() => setIsNewOpen(false)}
           componentProperties={componentProperties}
           savePackage={(x) => savePackage('new', x)}
+          packages={packages}
         />
       )}
       {isEditOpen && (
@@ -164,6 +165,7 @@ export const CalculationInput = ({
           componentProperties={componentProperties}
           editablePackage={selectedPackage}
           savePackage={(x) => savePackage('edit', x)}
+          packages={packages}
         />
       )}
     </>
--- a/web/src/components/feature/package_dialog/FluidDialog.tsx
+++ b/web/src/components/feature/package_dialog/FluidDialog.tsx
@@ -77,8 +77,6 @@ export const FluidDialog = ({
     [id: string]: boolean
   }>({})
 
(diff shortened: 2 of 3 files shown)


web/src/components/feature/calculation_input/CalculationInput.tsx
```diff
@@ -164,6 +165,7 @@ export const CalculationInput = ({
           componentProperties={componentProperties}
           editablePackage={selectedPackage}
           savePackage={(x) => savePackage('edit', x)}
+          packages={packages}
```
ingeridhellen, line 168: Does it make sense to be able to select a template when editing?

web/src/components/feature/package_dialog/FluidDialog.tsx
```diff
@@ -102,6 +112,17 @@ export const FluidDialog = ({
...
+              label="Template"
+              options={packages}
+              optionLabel={(option) => option.name}
+              onOptionsChange={(changes) => {
+                setComponentRatios(changes.selectedItems[0].components)
+                setSelectedPackage(changes.selectedItems[0])
+              }}
+              selectedOptions={selectedPackage ? [selectedPackage] : []}
+              autoWidth
+            />
```
ingeridhellen, line 125: We don't need this component to be controlled, so I suggest simplifying it
```suggestion
            <Autocomplete
              label="Template"
              options={packages}
              optionLabel={(option) => option.name}
              onOptionsChange={(changes) => {
                setComponentRatios(changes.selectedItems[0].components)
              }}
              autoWidth
            />
```
and then you can remove the selectedPackage state

web/src/components/feature/calculation_input/CalculationInput.tsx
soofstad: Was thinking about that as well. It might be handy in some cases (maybe?). And it doesn't really hurt.

web/src/components/feature/package_dialog/FluidDialog.tsx
soofstad: good point, fixed

ingeridhellen (approved):

Issue #183, opened 2022-11-04
feature(api): find a way to track usage statistics (how many calculations a month etc.)


soofstad:
Using metrics in radix might be a lightweight solution

soofstad:
This is done, but is currently running at the wrong location. "/api/v1/metrics" instead of "/metrics"

soofstad:
Standard total requests metrics are good enough for now

PR #193, opened 2022-11-07
feature(api): expose metrics

soofstad:

PO wants to be able to see usage statistics of the tool

Adds a `/metrics` endpoint in according to https://www.radix.equinor.com/guides/monitoring/#metrics-visualisation.

NOTE: Consider this a test/proof-of-concept.
This is a very primitive light weight solution, the "count" metric type that is now exported might not be a good fit.
Would like to get this into radix and see how it looks.

related to #183  but not closing it yet...

### DIFF
--- a/api/src/features/metrics/metrics_feature.py
+++ b/api/src/features/metrics/metrics_feature.py
@@ -1,0 +1,20 @@
+from fastapi import APIRouter
+from fastapi.responses import PlainTextResponse
+
+from config import config
+
+router = APIRouter(tags=["metrics"])
+
+
+class Metrics:
+    MULTIFLASH_CALCULATION_COUNT: int = 0
+    FETCH_COMPONENTS_COUNT: int = 0
+
+
+metrics = Metrics()
+
+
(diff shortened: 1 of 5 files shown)


api/src/features/metrics/metrics_feature.py
```diff
@@ -0,0 +1,20 @@
...
+    MULTIFLASH_CALCULATION_COUNT: int = 0
+    FETCH_COMPONENTS_COUNT: int = 0
+
+
+metrics = Metrics()
+
+
+@router.get("/metrics", response_class=PlainTextResponse, summary="Collect application metrics", response_model=str)
+async def get() -> str:
+    return f"""mercury_calculations_count {{env={config.ENVIRONMENT}}} {metrics.MULTIFLASH_CALCULATION_COUNT}\n
```
mariuswinger, line 19: Why triple quotes?

mariuswinger (approved):
Looks good.

api/src/features/metrics/metrics_feature.py
soofstad: For easier reading and writing of multi line strings.
Would be uglier to use many brackets and +'es IMO

PR #216, opened 2022-11-08
fix: update version reading

### LINKED ISSUE #215: bug(version): version message in dev does not look good

mariuswinger:
Version was looking weird.

Fixed version reading.

Closes #215


web/src/components/common/Header.tsx
```diff
@@ -27,15 +27,22 @@ export const Header = (): JSX.Element => {
     axios
       .get('version.txt')
       .then((response: AxiosResponse<string>) => {
-        const responseSubString = response.data.substring(
-          response.data.indexOf('tag:')
+        // get ref part from commit
+        const refSubString = response.data.substring(
+          response.data.lastIndexOf('('),
+          response.data.lastIndexOf(')') + 1
```
ingeridhellen, line 33: What if the commit message is `fix(dialog): Fixed the dialog`? Then you would get `refSubString == "dialog"`

.github/workflows/publish.yaml
```diff
@@ -78,7 +78,7 @@ jobs:
           echo "Tagging with ${{ inputs.image-tag }}"
           echo ${{ secrets.GITHUB_TOKEN }} | docker login $IMAGE_REGISTRY -u $GITHUB_ACTOR --password-stdin
           docker pull $API_IMAGE
-          printf "$(git log -n 1 --format=format:'%h% %>(12) %cs %aN%  %s %d')" > ./api/src/version.txt
+          printf "$(git log -n 1 --format=format:'%h% %>(12) %cs %aN%  "%s" %d')" > ./api/src/version.txt
```
ingeridhellen, line 81: Have you tried `git log -n 1 --format=format:'%d' --decorate-refs=refs/tags`. Then you would only get the tag

ingeridhellen (changes requested):

web/src/components/common/Header.tsx
mariuswinger: Nope, because refSubString extracts from reverse only the refs (which is always included, inside parentheses and always at the end of the string).

ingeridhellen: Right, you used `lastIndexOf`, then it would work, I suppose.

.github/workflows/publish.yaml
```diff
@@ -78,7 +78,7 @@ jobs:
           echo "Tagging with ${{ inputs.image-tag }}"
           echo ${{ secrets.GITHUB_TOKEN }} | docker login $IMAGE_REGISTRY -u $GITHUB_ACTOR --password-stdin
           docker pull $API_IMAGE
-          printf "$(git log -n 1 --format=format:'%h% %>(12) %cs %aN%  %s %d')" > ./api/src/version.txt
+          printf "$(git log -n 1 --format=format:'%h% %>(12) %cs %aN%  "%s"%d' --decorate=short --decorate-refs=refs/tags)" > ./api/src/version.txt
```
ingeridhellen, line 81: Why do you want to store the git hash, date, commit message and author when you don't use them?

mariuswinger: Don't really have a better answer than "that's how the template does it". @sooftsad can probably elaborate.

soofstad (changes requested):

mariuswinger:
Screenshot 2022-11-10 at 12 44 44
Screenshot 2022-11-10 at 12 45 00

Reformated the version.txt files and how it looks in web

web/src/components/common/Header.tsx
```diff
@@ -32,19 +36,35 @@ export const Header = (): JSX.Element => {
...
+          response.data.split('\n').map((line) => {
+            return line.split(': ')
+          })
         )
+        setVersionFile(versionFile)
       })
       .catch(() => null)
   }, [])
 
+  const VersionText = () => {
```
ingeridhellen, line 49: Could have been extracted into a separate file along with the useEffect

web/src/components/common/Header.tsx
```diff
@@ -32,19 +36,35 @@ export const Header = (): JSX.Element => {
...
+    if (versionFile.hash === '' || versionFile.date === '') {
+      return <p>Version: Null</p>
+    } else {
+      const link = (
+        <a
```
ingeridhellen, line 54: https://eds-storybook-react.azurewebsites.net/?path=/docs/typography-typography--link might be better

ingeridhellen (approved):
A few details, but nothing you need fix

web/src/components/common/Header.tsx
mariuswinger: Oh nice! Was looking for exactly this and did not know eds had it. Will implement.

soofstad (approved):

web/src/components/common/Header.tsx
```diff
@@ -32,19 +36,35 @@ export const Header = (): JSX.Element => {
...
+        setVersionFile(versionFile)
       })
       .catch(() => null)
   }, [])
 
+  const VersionText = () => {
+    if (versionFile.hash === '' || versionFile.date === '') {
+      return <p>Version: Null</p>
+    } else {
+      const link = (
```
soofstad, line 53: Why declare this "link" var before returning it? :thinking:

soofstad (approved):

web/src/components/common/Header.tsx
mariuswinger: Fixed

PR #220, opened 2022-11-08
fix: Fix component table order

### LINKED ISSUE #201: feature(web): order of components table should be defined by the order they were added

ingeridhellen:

Eleni wants the order of the components table to be determined by the order they were added

Fixed by switching TComponentProperties from object to array

Closes #201

### DIFF
--- a/web/src/components/feature/package_dialog/ComponentSelector.tsx
+++ b/web/src/components/feature/package_dialog/ComponentSelector.tsx
@@ -13,13 +13,6 @@ const ComponentSelectorContainer = styled.div`
   max-width: 420px;
 `
 
-type TComponentProperty = {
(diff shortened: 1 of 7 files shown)


soofstad (approved):
Looks good, but the order is not changed in the API yet, which is needed for this to work right?

mariuswinger (approved):

mariuswinger:
> Looks good, but the order is not changed in the API yet, which is needed for this to work right?

Checked. Did not work. The components are somehow sorted during transit. Very weird.

ingeridhellen:
This issue is not related to the order of the components or the API. It is related to keeping the order in the table when adding new components to the table

PR #226, opened 2022-11-08
fix: Avoid both popovers open at once

ingeridhellen:

Avoid both popovers being open at once, causing one to cover the other

Automatically closes the other popover when one of the popover opening buttons are clicked


soofstad (approved):
Thats sane behaviour :)

PR #227, opened 2022-11-09
feat: no more zero mercury packages

### LINKED ISSUE #178: feat(web): Warning when trying to create package with 0 mercury

mariuswinger:
Packages can not have zero mercury

Not allowed to save packages with zero mercury.

Closes #178

### DIFF
--- a/web/src/components/feature/package_dialog/ComponentTable.tsx
+++ b/web/src/components/feature/package_dialog/ComponentTable.tsx
@@ -24,6 +24,24 @@ export const ComponentTable = ({
   ratiosAreValid: { [id: string]: boolean }
   setRatiosAreValid: (x: { [id: string]: boolean }) => void
 }): JSX.Element => {
+  function handleOnChange(
+    event: React.ChangeEvent<HTMLInputElement>,
+    id: string
+  ) {
+    const ratio = event.target.value
+    const newRatiosAreValid = { ...ratiosAreValid }
+    newRatiosAreValid[id] =
+      event.target.checkValidity() && !isNaN(Number(ratio))
+    setRatiosAreValid(newRatiosAreValid)
+    const newRatios = { ...componentRatios }
+    if (ratio === '') {
+      delete newRatios[id]
+    } else {
+      newRatios[id] = ratio
+    }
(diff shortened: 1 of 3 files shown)


ingeridhellen (approved):

web/src/components/feature/package_dialog/ComponentTable.tsx
```diff
@@ -36,23 +54,14 @@ export const ComponentTable = ({
             value={componentRatios[component.id] ?? ''}
             pattern="^\d+(\.\d+)?([eE][-+]?\d+)?$"
             variant={
-              ratiosAreValid[component.id] === false ? 'warning' : undefined
+              ratiosAreValid[component.id] === false ||
+              (component.id === '5' && !(Number(componentRatios['5']) > 0))
```
soofstad, line 58: But I would rather have this special case logic inside "handleChange".
It's inside that function we set if ratios are valid or not.

They way it is written now. You can not trust "ratiosAreValid" and need to do extra checks several places

web/src/components/feature/package_dialog/SaveButton.tsx
```diff
@@ -34,7 +34,8 @@ export const SaveButton = (props: {
     const isValidPackage =
       props.packageName.length > 0 &&
       Object.values(props.ratiosAreValid).every((x) => x) &&
-      Object.values(props.componentRatios).some((x) => Number(x) > 0)
+      Object.values(props.componentRatios).some((x) => Number(x) > 0) &&
+      Number(props.componentRatios['5']) > 0
```
soofstad, line 38: Same with this. Should be enough to check "ratiosAreValid"

soofstad (changes requested):

web/src/components/feature/package_dialog/ComponentTable.tsx
ingeridhellen: Perhaps renaming ratiosAreValid to inputsAreValidNumbers then. The solution you propose is more complicated, since it means we would need to add extra logic to the onOptionsChange for the autocomplete in ComponentSelector.

soofstad (approved):

PR #228, opened 2022-11-09
Feat/component order

### LINKED ISSUE #191: Feat(web): Order components to start with Mercury, H2S, Water, MEG, TEG, and rest....

ingeridhellen:

This pull request is needed because Eleni wants components to be ordered as [Mercury, H2S, Water, Meg, Teg, ..rest]

- Change TComponentProperties to TComponentProperty[] (necessary to get typing easily)
- Add sorting upon getting the components from the API
- Change how mole table gets its rows to ensure correct order there as well

Closes #191

### DIFF
--- a/web/src/pages/Main.tsx
+++ b/web/src/pages/Main.tsx
@@ -9,7 +9,7 @@ import { AxiosResponse } from 'axios'
 import { ComponentResponse } from '../api/generated'
 import { PhaseTable } from '../components/feature/results/PhaseTable'
 import { MercuryWarning } from '../components/feature/results/MercuryWarning'
-import { TComponentProperties, TResults } from '../types'
+import { TComponentProperty, TResults } from '../types'
 
 const Results = styled.div`
   display: flex;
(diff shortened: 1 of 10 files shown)


web/src/pages/Main.tsx
```diff
@@ -32,10 +33,18 @@ const DividerWithLargeSpacings = styled(Divider)`
   margin-top: 40px;
 `
 
+const sortComponents = (components: TComponentProperty[]) => {
+  const firsts = orderedComponents
+    .map((id) => components.find((x) => x.id === id))
+    .filter((x): x is TComponentProperty => !!x)
```
mariuswinger, line 39: No biggie, but if we are going to use find/lookup anyways, couldn't we just pass response.data.components directly which is a dict/object and create the list/TComponentProperty[] inside sortComponents using the custom order?

mariuswinger (approved):

web/src/pages/Main.tsx
ingeridhellen: You're right, this was better

PR #243, opened 2022-11-15
refactor: Remove setAreRatioValid

ingeridhellen:

This pull request is needed because of....

Removes the logic of validating ratios into the reducer


soofstad (approved):
Good!

PR #245, opened 2022-11-15
ci: Fix release workflow

### LINKED ISSUE #244: CI: Deploy to prod

ingeridhellen:

We want to move to prod

- Triggers on release tags on main only
- Fix bugs in workflow file
- Adds release tag to github image

Closes #244

### DIFF
--- a/.github/workflows/on-tag.yaml
+++ b/.github/workflows/on-tag.yaml
@@ -7,11 +7,13 @@ on:
 
 jobs:
   tests:
+    if: github.ref == 'refs/heads/main'
     uses: ./.github/workflows/tests.yaml
     secrets:
       LIBHG_REPO_DEPLOY_KEY: ${{ secrets.LIBHG_REPO_DEPLOY_KEY }}
 
   set-tag-ref:
+    if: github.ref == 'refs/heads/main'
     runs-on: ubuntu-latest
     outputs:
       tag-ref: ${{ steps.get-tag-ref.outputs.tag-ref }}
--- a/.github/workflows/on-tag.yaml
+++ b/.github/workflows/on-tag.yaml
@@ -1,9 +1,9 @@
-name: "New tag"
+name: "On tag: release main to prod"
 on:
   workflow_dispatch:
   push:
     tags:
-      - v*
+      - "v[0-9]+.[0-9]+.[0-9]+"
 
 jobs:
   tests:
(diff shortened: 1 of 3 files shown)


.github/workflows/publish.yaml
```diff
@@ -1,27 +1,31 @@
...
         required: true
         type: string
       oauth-redirect-url:
-        description: 'Redirect url for oauth. Should be the public url to access the web app'
-        default: 'https://proxy-mercury-dev.radix.equinor.com'
+        description: "Redirect url for oauth. Should be the public url to access the web app"
+        default: "https://proxy-mercury-dev.radix.equinor.com"
         required: true
         type: string
+      release-tag:
```
soofstad, line 16: I think it would be cleaner if the workflow takes a "tags" array as input, and after building the image, runs over that array in parallel to tag and upload images.

If that's is not possible, calling the workflow twice with a single tag also helps to keep the workflows more "functional"

.github/workflows/publish.yaml
```diff
@@ -66,7 +72,9 @@ jobs:
           --tag ${WEB_IMAGE} \
           ./web
           docker tag $WEB_IMAGE $WEB_IMAGE:${{ inputs.image-tag }}
+          if [ -n "${{ inputs.release-tag }}" ]; then docker tag $WEB_IMAGE $WEB_IMAGE:${{ inputs.release-tag }}; fi;
           docker push $WEB_IMAGE:${{ inputs.image-tag }}
+          if [ -n "${{ inputs.release-tag }}" ]; then docker push $WEB_IMAGE:${{ inputs.release-tag }}; fi;
 
   build-and-publish-api-master:
```
soofstad, line 79: Not your change; but these names are misleading. There is nothing in this workflow indicating "master".

soofstad (changes requested):

.github/workflows/publish.yaml
ingeridhellen: Github actions only support inputs of type boolean, number or string. I could take in a comma separated string and split it up though.
If we call the workflow twice (that was what it did before) I think we'll create the image twice too. Now we create one image with two tags

.github/workflows/on-tag.yaml
```diff
@@ -1,46 +1,41 @@
-name: "New tag"
+name: "On tag: release main to prod"
 on:
   workflow_dispatch:
   push:
```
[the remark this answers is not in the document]
ingeridhellen: Thought about that. There is not a very clean way of doing that, but could do as in https://github.com/equinor/NeqSimAPI/commit/a81ffe44ded31bbbef3a2b5e593ced3a37e70e53

.github/workflows/publish.yaml
ingeridhellen: And how would you do the tagging in parallel? Jobs run in parallel, but they don't share environment, so I would have to push the image with one of the tags only to pull it down again and retag it in the next workflow.

soofstad (approved):

PR #248, opened 2022-11-16
ci: Fix bug in publish workflow

ingeridhellen:

Bug in https://github.com/equinor/mercury/pull/245

Fixed how image-tags was parsed

### DIFF
--- a/.github/workflows/publish.yaml
+++ b/.github/workflows/publish.yaml
@@ -44,7 +44,7 @@ jobs:
           --tag ${NGINX_IMAGE} \
           ./nginx
           IFS=','
-          for IMAGE_TAG in `echo ${{ inputs.image-tags }}`
+          for IMAGE_TAG in $(echo ${{ inputs.image-tags }})
           do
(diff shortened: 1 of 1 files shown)


.github/workflows/publish.yaml
```diff
@@ -44,7 +44,7 @@ jobs:
           --tag ${NGINX_IMAGE} \
           ./nginx
           IFS=','
-          for IMAGE_TAG in {${{ inputs.image-tags }}}
+          for IMAGE_TAG in `echo ${{ inputs.image-tags }}`
```
soofstad, line 47: Is this really necessary? :thinking: 
The outer most brackets from before should be removed, and that's that.
Like; `for IMAGE_TAG in  ${{ inputs.image-tags }}; do`

ingeridhellen: Tried that, did not work. It wouldn't split the image-tags before looping

soofstad (approved):

PR #286, opened 2022-12-05
feature: create an ErrorBoundary

### LINKED ISSUE #285: feature: add error boundary

soofstad:
- Need an error boundary to inform users how to get out of a bad application state/handle errors

- Adds info on technical support email
- Adds one top level error boundary
Screenshot_20221205_094925

closes #285

### DIFF
--- a/web/src/common/ErrorBoundary.tsx
+++ b/web/src/common/ErrorBoundary.tsx
@@ -12,7 +12,7 @@ interface State {
 
 class ErrorBoundary extends Component<Props, State> {
   public state: State = {
-    hasError: false,
+    hasError: true,
   }
 
   public static getDerivedStateFromError(_: Error): State {
@@ -36,7 +36,7 @@ class ErrorBoundary extends Component<Props, State> {
           </Typography>
           <Typography varant={'body_short'}>
             This will cause you to lose application data. If that is not a good
-            option you should contact support.
+            option you should contact support (team-hermes@equinor.com).
           </Typography>
         </>
       )
--- a/web/src/common/ErrorBoundary.tsx
+++ b/web/src/common/ErrorBoundary.tsx
@@ -1,5 +1,5 @@
 import { Component, ErrorInfo, ReactNode } from 'react'
-import { Typography, Icon } from '@equinor/eds-core-react'
+import { Typography, Icon, List } from '@equinor/eds-core-react'
 import { account_circle } from '@equinor/eds-icons'
 
 interface Props {
@@ -12,7 +12,7 @@ interface State {
 
 class ErrorBoundary extends Component<Props, State> {
   public state: State = {
-    hasError: true,
+    hasError: false,
   }
 
   public static getDerivedStateFromError(_: Error): State {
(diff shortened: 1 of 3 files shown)


web/src/common/Header.tsx
```diff
@@ -107,6 +107,7 @@ export const Header = (): JSX.Element => {
         <Popover.Content>
           <VersionText />
           <p>Person of contact: Eleni Pantelli (elp@equinor.com)</p>
+          <p>Technical Support: team-hermes@equinor.com</p>
```
ingeridhellen, line 110: Have this email been tested? I expected a test email, but I never got one

web/src/common/ErrorBoundary.tsx
```diff
@@ -0,0 +1,49 @@
...
+          <Typography variant={'h3'}>Ops... Something went wrong 😞</Typography>
+          <Typography varant={'body_short'}>
+            You could try resetting the application by clicking the{' '}
+            <Icon data={account_circle} /> icon, and then &quot;reset
+            application data&quot;.
+          </Typography>
+          <Typography varant={'body_short'}>
+            This will cause you to lose application data. If that is not a good
+            option you should contact support (team-hermes@equinor.com).
+          </Typography>
```
ingeridhellen, line 40: Sorry, I know I said this was good, but now I am not so sure. Many times, an error may be triggered by some odd combination of commands that is unlikely to happen twice in a row (ex: selecting a template, then deselecting it again, this caused the app to crash earlier - there are no reason why anyone would do that twice). The first thing people should do if something fails is therefore simply to reload the page and try again. The text as it is may cause people to lose data when they don't really have to.

web/src/common/ErrorBoundary.tsx
```diff
@@ -0,0 +1,49 @@
...
+
+interface Props {
+  children?: ReactNode
+}
+
+interface State {
+  hasError: boolean
+}
+
+class ErrorBoundary extends Component<Props, State> {
```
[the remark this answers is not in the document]
ingeridhellen: Ah, I see, it can't be done otherwise at the moment. There is a [npm package](https://github.com/bvaughn/react-error-boundary) that could be used, but it doesn't seem to be properly maintained, so I suppose this is the best option.

web/src/common/ErrorBoundary.tsx
```diff
@@ -0,0 +1,49 @@
...
+  public state: State = {
+    hasError: true,
```
ingeridhellen, line 15: Ops

ingeridhellen (changes requested):
A lot of nice stuff here

web/src/common/ErrorBoundary.tsx
soofstad: Rewrote into a numbered list, where 1. is "Reload page"

web/src/common/ErrorBoundary.tsx
soofstad: fixed!

soofstad:
Updated text
Screenshot_20221205_140723

web/src/common/Header.tsx
soofstad: updated and tested ok

ingeridhellen (approved):

PR #300, opened 2022-12-13
fix: optional refresh when auth session expires

### LINKED ISSUE #299: web(auth): Add onRefreshTokenExpire

soofstad:

- The user should not have the page refresh without any warning/message

- Adds a "confirm/deny" native dialog to the "onSessionExpire" event listener.
- Bump auth package
- Move unnecessary variables from web Dockerfile to source code 

Write summary of what this pull request changes if needed.

closes #299


web/src/constants.ts
```diff
@@ -1,12 +1,18 @@
...
-  tokenEndpoint: process.env.REACT_APP_TOKEN_ENDPOINT || '',
-  scope: process.env.REACT_APP_AUTH_SCOPE || '',
+  clientId: '0cd10735-a40c-4ff6-84c8-74e3b54ed93f',
+  authorizationEndpoint:
+    'https://login.microsoftonline.com/3aa4a235-b6e2-48d5-9195-7fcf05b459b0/oauth2/v2.0/authorize',
+  tokenEndpoint:
+    'https://login.microsoftonline.com/3aa4a235-b6e2-48d5-9195-7fcf05b459b0/oauth2/v2.0/token',
+  scope: 'api://0cd10735-a40c-4ff6-84c8-74e3b54ed93f/All',
   redirectUri: process.env.REACT_APP_AUTH_REDIRECT_URI || '',
-  logoutEndpoint: process.env.REACT_APP_LOGOUT_ENDPOINT || '',
```
ingeridhellen, line 9: We have a logout button. Will it still function if you remove the logoutEndpoint?

soofstad: Yes!
Except that it will only log you out of the Mercury app (clear local auth state).
Instead of logging out of the Azure AD single-sign-on session, which is often not what you want or expect.

ingeridhellen: Then it sounds like an improvement!

ingeridhellen (approved):

PR #314, opened 2023-01-02
chore: Fix gitignore bug

ingeridhellen:

This pull request is needed because our gitignore file contains entries that excludes the yarn lib folders. This causes problems for new runners of the code

Similar to https://github.com/equinor/template-fastapi-react/commit/194018683344dfaf9e982474b2024f4f6d8b099f#diff-a5de3e5871ffcc383a2294845bd3df25d3eeff6c29ad46e3a396577c413bf357, it ensures the lib folders are checked in.


.gitignore
```diff
@@ -145,3 +145,4 @@ web/.yarn/*
 !web/.yarn/releases
 !web/.yarn/sdks
 !web/.yarn/versions
+!web/.yarn/**/lib/
```
soofstad, line 148: According to the yarn docs (https://next.yarnpkg.com/getting-started/qa#which-files-should-be-gitignored), our yarn related gitignore is correct.
This could also have unwanted effects, like adding a file named "lib" from the `.yarn/cache folder`

IMO it's a better fix to remove line 18 ("lib").
Noted as another problem with using huge auto-generated gitignore templates...

soofstad (changes requested):

soofstad (approved):

PR #320, opened 2023-01-23
fix: component result order

### LINKED ISSUE #319: feature: Keep order of components as defined in input when displaying result

ingeridhellen:

This pull request is needed because Eleni wants the components in the result to be ordered same as the components in the fluid package

Refactors TResults to use arrays instead of object, thereby enabling it to preserve order. Then adding order functionality.

Closes #319


soofstad (approved):
Nice! Testes locally OK. Good work :)

PR #351, opened 2023-02-03
Feat/last input context

### LINKED ISSUE #323: bug: Mercury phase marked as "liquid"
>I have noticed one last thing. At very low temperatures appr. -40oC and below, if a separate mercury phase is produced then this is solid not liquid. Now the tool says it is liquid. Can you please check and update accordingly? The relevant information about creation of solid phase should be available in the code.

mariuswinger:
Status banner needs to know input temperature to give correct state (liquid/solid) of mercury phase (if present).

Adds a context for the last input entered.

Closes #323

### DIFF
--- a/web/src/pages/Main.tsx
+++ b/web/src/pages/Main.tsx
@@ -12,6 +12,7 @@ import { TCalcStatus, TComponentProperty, TResults } from '../types'
 import { orderedComponents } from '../constants'
 import { Status } from '../feature/Status'
 import ErrorBoundary from '../common/ErrorBoundary'
+import { LastInputProvider } from './context/LastInputContext'
 
 import { useAppInsightsContext } from '@microsoft/applicationinsights-react-js'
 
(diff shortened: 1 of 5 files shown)


web/src/pages/Main.tsx
```diff
@@ -50,6 +51,8 @@ export const MainPage = (props: { mercuryApi: MercuryAPI }): JSX.Element => {
   const [isLoading, setIsLoading] = useState<boolean>(true)
   const [calcStatus, setCalcStatus] = useState<TCalcStatus>()
   const [result, setResult] = useState<TResults>()
+  // const [lastCalculation, setLastCalculation] = useState<TCalculation>({})
+  // const [nextCalculation, setNextCalculation] = useState<TCalculation>(null)
```
ingeridhellen, line 55: remove these

soofstad (approved):
Great! Very clean

web/src/pages/context/LastInputContext.tsx
```diff
@@ -0,0 +1,34 @@
+import { createContext, useContext, useState } from 'react'
+import { Multiflash } from '../../api/generated'
+
+const LastInputContext = createContext<
+  | { lastInput: Multiflash; setLastInput: (input: Multiflash) => void }
+  | undefined
+>(undefined)
```
ingeridhellen, line 7: Why that name and these types? Currently, only temperature is used. IMO it would make more sense to remove pressure and composition from this context, and instead put the data from TResult into it. Perhaps rename LastInputContext to MainPageContext and TResults to TMainPage, and extend TMainPage to include temperature?

soofstad: I imaged we would use this context, along with a "resultContext", in future refactors.
So having the whole "lastInput" object gives it more potential for the future.

soofstad (approved):

PR #407, opened 2023-05-03
chore: bump dependencies in api and web

mariuswinger:
Bumping of dependencies.


ingeridhellen:
What is the purpose of the files in api/src/info? It doesn't seem to be related to the dependency bumping

mariuswinger:
> What is the purpose of the files in api/src/info? It doesn't seem to be related to the dependency bumping

True. Will remove.

PR #411, opened 2023-05-04
chore: bump remaining packages

mariuswinger:
Dependabot screwed up. Had to update some packages manually.


ingeridhellen (approved):
Are you familiar with the `yarn upgrade-interactive` command? It is quite handy in these cases

mariuswinger:
> Are you familiar with the `yarn upgrade-interactive` command? It is quite handy in these cases

Yes, I used it. But it was not able to discover webpack, since that is a dependency from react that is not in our package.json file. Also, it was not able to discover the newest package of applicationinsights at the time when I bumped the other packages (even though it had a newer version than the one displayed).

soofstad (approved):

PR #452, opened 2023-09-05
chore: remove PnP and upgrade web dependencies

ingeridhellen:

New packages are better. And yarn pnp is difficult at best

Removes yarn pnp and updates all web dependencies except
- @typescript-eslint/eslint-plugin
- @typescript-eslint/parser
- eslint-plugin-react
- axios

The three first causes eslint to fail, while the latter causes the tests to fail. I think the former is related to us using CRA, which uses old versions of these packages, and causes eslint to be confused as to which to use.


mariuswinger (approved):
All for it

PR #617, opened 2025-02-27
fix: replace opencensus and use exception handling routes to export telemetry

collinlokken:

OpenCensus is deprecated and we should migrate to OpenTelemetry

Replace OpenCensus with OpenTelemetry and standardise exception handling for easy telemetry export


MarcusDragsten (approved):
Good stuff! ✅

api/src/entities/ComponentResponse.py
```diff
@@ -28,9 +28,9 @@ class ComponentResponse(BaseModel):
         description="Dictionary of component_ids as string and ComponentProperties as value",
     )
 
-    class Config:
-        allow_mutation = False
-        schema_extra = {
+    model_config = ConfigDict(
+        frozen=False,
```
mariuswinger, line 32: Not sure, but it sounds like `allow_mutation = False` is not the same as `frozen = False`?

mariuswinger (approved):
How was exceptions handled previously?

api/src/entities/ComponentResponse.py
collinlokken: `api/src/common/responses.py`

mariuswinger: Ah, I see, it was a decorator. This is way better.

collinlokken: `frozen` should be `True`, correct

PR #621, opened 2025-02-27
fix: install poetry without dev

collinlokken:

CI unable to use --no-dev flag

replace with --only=main


MarcusDragsten (approved):
🧑‍🏭

PR #624, opened 2025-02-28
ci: add federated credentials for radix deploy

collinlokken:

Mercury client secret expired recently. This PR moves to federated credentials for deploy instead of client secret.

Replace client secret in radix deploy with github federated credentials


soofstad (commented):
Need: permissions:
27   │   id-token: write
28   │   contents: read

and can remove the secret SERVICE_PRINCIPAL_CLIENT_SECRET

soofstad (approved):

PR #717, opened 2026-03-02
chore: Add  to field validator

einarwar:
POOL WEEK

fixes CodeQL issue
image

Adds `@classmethod` decorator as described by [pydantic docs](https://docs.pydantic.dev/latest/concepts/validators/#field-validators)


mariuswinger (commented):
Good, but we can also add typing I think

Issue #697, opened 2026-03-02
Update pre-commit


mariuswinger:
- [x] check if hooks need updating
 - [x] check if some hooks are missing
 - [x] consider using local biome and ruff to fix versions to project versions

PR #721, opened 2026-03-03
ops: pre commit local ruff biome

mariuswinger:
⚠️  Branches onto #718. Will probably fail in actions until #719 is merged.

POOL WEEK.

Makes it so pre-commit uses the locally installed versions of ruff and biome. No more having to update versions several places.

#697


einarwar (approved):
Love it!

PR #743, opened 2026-03-05
fix(web): handle unauthorized error on component fetch

### LINKED ISSUE #733: fix: handle unauthorized exception better
#### Expected Behaviour

Errors should be handled appropriately.

#### Actual Behaviour

Error is not handled appropriately. Due to useEffect and conditional rendering in `MainPage.tsx`

#### Steps to Reproduce

If token has expired, the web page does not render simply outputting "Failed to fetch list of components" and user must manually refresh page to get a new token.

einarwar:
## Summary

Properly handles errors when fetching the component list, particularly unauthorized (401) errors from expired tokens.

## Problem

When the user's token expires, `ComponentService.getComponents()` fails silently (no `.catch()` handler). The page shows a dead-end "Failed to fetch list of components" message with no way to recover except manually refreshing the browser.

## Changes

In `web/src/pages/MainPage/MainPage.tsx`:
- Added error state tracking for the fetch call
- Added `.catch()` handler to capture errors from `getComponents()`
- For **401 errors**: shows "Your session may have expired" with a **Log in again** button
- For **other errors**: shows a generic error message with a **Retry** button
- Extracted fetch logic into a `useCallback` so it can be retried

In `web/src/pages/MainPage/MainPage.test.tsx`:
- Added test for retry button on generic fetch failure
- Added test for login button on 401 unauthorized error

Closes #733

### DIFF
--- a/web/src/components/ErrorState/ErrorState.tsx
+++ b/web/src/components/ErrorState/ErrorState.tsx
@@ -1,0 +1,32 @@
+import { Button, Typography } from '@equinor/eds-core-react'
+import type { IconData } from '@equinor/eds-icons'
+import { Flexbox } from '../Flexbox/Flexbox'
+import { ErrorIcon, Wrapper } from './styles'
+
+type ErrorStateProps = {
+  icon: IconData
+  title: string
+  message: string
+  buttonLabel: string
+  onAction: () => void
+}
+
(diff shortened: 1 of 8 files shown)


matsklevstad (approved):
Works as expected, nice ✅ If I were to change anything, remove “may” because the session has definitely expired. But only nitpicking!

